PR #45, opened 2019-12-07
Add array literal support

as3810t:
This pull request aims to add basic array literal support to theta.


subprojects/solver-z3/src/test/java/hu/bme/mit/theta/solver/z3/Z3SolverTest.java
```diff
@@ -105,7 +110,26 @@ public void testFunc() {
...
+        solver.add(ArrayExprs.Eq(Write(arr.getRef(), Int(1), Int(2)), arr.getRef()));
+
+        // Check, the expression should be satisfiable
+        SolverStatus status = solver.check();
+        assertTrue(status.isSat());
+        
+        Valuation valuation = solver.getModel();
+        final Optional<LitExpr<ArrayType<IntType, IntType>>> optVal = valuation.eval(arr);
+		final Expr<ArrayType<IntType, IntType>> val = optVal.get();
+		assertEquals(arr.getType(), val.getType());
```
hajduakos, line 132: As far as I see, this only tests that the types match (Int index and Int values). Could you add assertions to test the literals inside the array, i.e., is element 0 really Int(1) and is element 1 really Int(2)?

hajduakos:
The FuncInterp corresponding to the array can have a default branch in general what happens to that? Is it filled with some value? Maybe we should put that in the array literal.

PR #49, opened 2020-05-21
Xsts

mondokm:


hajduakos:
@mondokm please write some tests in `core/utils/StmtToExprTransformerTest.java`. There are tests for assume/assert/havoc, but it would be nice to have tests for the new statements (seq/ort/nondet)

PR #59, opened 2020-07-28
Xcfa partial order

rl555:
Merge changes from the master branch.
Add an auxiliary transformations to XCFAs (split edges with multiple stmts), add possibility to have a single XCFA split among multiple files (multiple process declarations are seen as one).
Add a way to convert XCFA to CFA (a bit more powerful than the one provided in XCFA class)

Note that due to a a problematic rebase there are unrelated changes here :/


rl555:
All important changes are present at radl97/theta
Sorry for the too much branches/forks

PR #69, opened 2020-08-26
Replace int as underlying type in IntLitExpr and RatLitExpr

### LINKED ISSUE #42: IntLitExpr should have a larger value range
Programs cannot be verified if they contain integers (such as 4294967295) which are larger than the limit of the signed java int type. The internal value held by IntLitExpr should be replaced with long (at least for the time being). This may also affect certain DSL parsers.

as3810t:
This pull request replaces int as underlying type in IntLitExpr and RatLitExpr, subsequently closes #42 .


as3810t:
As of now, theta builds successfully, however more integration tests should be run to evaluate the changes.

subprojects/core/src/test/java/hu/bme/mit/theta/core/utils/ExprSimplifierTest.java
```diff
@@ -269,6 +270,7 @@ public void testIntAdd() {
 		assertEquals(Add(a, Int(4)), simplify(Add(Int(1), Add(a, Int(3)))));
 		assertEquals(a, simplify(Add(Int(-3), a, Int(3))));
 		assertEquals(Add(a, b, a, b, c), simplify(Add(a, Add(b, Add(a, Add(b, c))))));
+		assertEquals(Int(new BigInteger("4294967294")), simplify(Add(Int(Integer.MAX_VALUE), Int(Integer.MAX_VALUE))));
```
hajduakos, line 273: Nice

hajduakos (approved):
Looks good

hajduakos:
I'm just about to run tests. I'll get back when there are results

subprojects/core/src/test/java/hu/bme/mit/theta/core/utils/ExprSimplifierTest.java
as3810t: Right now `Int` takes either `int` or a `BigInteger`. What do you think, should I implement az overload that takes a `String`, making it easier to create "big" literals?

hajduakos: Yes, I think that would be useful. Basically that would just pass `new BigInteger(string)` to the existing constructor, right?

hajduakos: By constructor I mean the static factory method.

as3810t: Implemented in b05cd37

PR #77, opened 2020-09-05
Remove signedness info from bitvector type

as3810t:
Bitvectors can be interpreted either as an unsigned number or as a signed number. There are operations where it does not matter due to the two's complement semantics (addition, subtraction, multiplication, etc.), where it does not matter due to the semantics of the operation (bitwise and, bitwise or, shifting, etc.), and where the operation is different based on the interpretation of the operands (division, remainder, etc.).

In the earlier versions of Theta, the bitvector type carried the information whether the literal, or operation should be treated as an expression yielding unsigned or signed value. This design decision was made to make the frontend languages more userfriendly.

However, the underlying SMT solvers, and the currently supported (LLVM based) C frontend, [GAZER](https://github.com/ftsrg/gazer) both treat their respective bitvector type as signless, and the information about the signs is inserted by using different operators for unsigned and signed operations where necessary.

This pull request refactors the current bitvector support, to remove the signedness information from the bitvector type, making the integration with both the frontend and backend more straightforward, less error-prone.


hajduakos (approved):
Fixed a typo in an operator and added a remark in the docs. Otherwise looks good!

PR #82, opened 2020-09-09
Advanced refinement algorithms supporting bitvectors

as3810t:
This pull request adds additional refinement strategies that can be used when interpolation is unavailable (for example in case of bitvector problems using Z3). The refinement strategies added are:
- UCB
- NEWTON_SP
- NEWTON_WP
- NEWTON_IT_SP
- NEWTON_IT_WP
- NEWTON_SP_LV
- NEWTON_WP_LV
- NEWTON_IT_SP_LV
- NEWTON_IT_WP_LV


hajduakos:
> 
> 
> My benchmarking showed that actually `NWT_IT_WP` is the most effective (in terms of that it can solve the most problems).

@as3810t okay, I will update the readme.

hajduakos (approved):
Nice work

Issue #29, opened 2017-11-03
Property description for CFAs


tothtamas28:

hajduakos:
The algorithms support arbitrary target predicates over states. However, the CFA tool always checks the reachability of the designated error location. It could be generalized by developing a simple property specification language for the CFA tool. For example:

- `--model="sample.cfa" --property="reach={L12}"` 
- `--model="sample.cfa" --property="expr={x > 0}"`

hajduakos:
Let's start with the first feature, namely being able to provide the error location separately (currently it is hardcoded in the CFA). For backward compatibility, I would still allow the error location to be specified inside the CFA, however, it would be optional, and command line tools would accept a `--errorloc <LOC NAME>` argument. There would be 4 cases:

|  | Error loc given in CFA | No error given loc in CFA |
|--|--|--|
| **Error loc given as argument** | Use argument, but possibly give a warning | Use argument |
| **No error loc given as argument** | Use error loc from CFA | Error: nothing to verify |

Any thoughts @vincemolnar @as3810t ?

vincemolnar:
Seems good to me, but maybe it would be better to give the argument a less specific (and tehrefore more futureproof) name, for example `--targetloc` (and then the result could be written more precisely too, e.g. `Verification completed, [target|error] location is [reachable|unreachable].`

hajduakos:
We have now support for arbitrary error locations with the `--errorloc` parameter. Supporting properties like `x >= 0` seems to have one critical point, which is large block encoding (LBE). In LBE, sequential paths (or sometimes even branching statements) are merged into a single step. This is not really possible if we want arbitrary properties. For example, `x--; x++;` could cause the property `x >= 0` to be violated in intermediate steps that would not be found with LBE.

PR #85, opened 2020-09-13
Make error loc of CFA optional, add targetloc argument to CLI tool

hajduakos:
This PR makes the final/error locations of the CFA optional. The motivation behind this is that (1) not all algorithms are interested in them, and (2) sometimes it can be desirable to check multiple queries for the same CFA. The CLI tool was extended in a way that it can accept a `--targetloc` argument. If the argument is given, that location is treated as error, otherwise we fallback to the old solution, namely looking for the error loc in the CFA. This makes the solution backward compatible. See #29 for more information

@as3810t Can you play around a bit with this version?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ buildscript {
 
 allprojects {
     group = "hu.bme.mit.inf.theta"
-    version = "2.6.1"
+    version = "2.7.0"
 
     apply(from = rootDir.resolve("gradle/shared-with-buildSrc/mirrors.gradle.kts"))
 }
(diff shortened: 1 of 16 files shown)


as3810t (commented):
It seems that everything works perfectly.

My one nitpick is, that right now `--targetloc` defines a custom error location. However it could also be used to check, whether a custom, not necessarily error location is reachable. In the long term it would be beneficial that when `--targetloc` is present, the output is either "target loc reachable (with cex)" or "target loc unreachable", instead of current safe / unsafe output.

hajduakos:
@as3810t as we discussed, I renamed `--targetloc` to `--errorloc` as currently the terminology of the algorithms and internal APIs revolve around error locations and safety.

PR #93, opened 2020-10-06
Implement reset in Z3 solver

### LINKED ISSUE #92: Support for reset in Z3 solver
Message is null.
Solver was Z3.

hajduakos:
This PR implements the reset function in the Z3 solver. While it is recommended to instead use push/pop, it can be still useful for quick prototyping. Furthermore, some messages were added to exceptions.

Fixes #92

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ buildscript {
 
 allprojects {
     group = "hu.bme.mit.inf.theta"
-    version = "2.7.0"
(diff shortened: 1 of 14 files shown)


as3810t (approved):
I think it works as intended. Maybe there could be another test case for the `reset` function similar to `testClear` in `Z3SolverTest` that checks whether `reset` resets the stack of the solver as well.

PR #96, opened 2020-10-26
Upgrade to Gradle 6.x

### LINKED ISSUE #95: Build fails on archlinux
Hi,

when I try to build theta I get the following error:

```
$ gradle build

Welcome to Gradle 6.6.1!

Here are the highlights of this release:
 - Experimental build configuration caching
 - Built-in conventions for handling credentials
 - Java compilation supports --release flag

For more details see https://docs.gradle.org/6.6.1/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)

> Task :buildSrc:compileKotlin
w: /home/marek/src/theta/buildSrc/src/main/kotlin/tool-common.gradle.kts: (22, 48): 'getter for version: String?' is deprecated. Deprecated in Java

> Configure project :
e: /home/marek/src/theta/build.gradle.kts:37:9: Val cannot be reassigned
e: /home/marek/src/theta/build.gradle.kts:38:9: Val cannot be reassigned
e: /home/marek/src/theta/build.gradle.kts:42:9: Val cannot be reassigned
e: /home/marek/src/theta/build.gradle.kts:45:29: Type mismatch: inferred type is FileCollection! but ConfigurableFileCollection! was expected

FAILURE: Build failed with an exception.

* Where:
Build file '/home/marek/src/theta/build.gradle.kts' line: 37

* What went wrong:
Script compilation errors:

  Line 37:         sourceDirectories = files(reportTasks.map { it.allSourceDirs })
                   ^ Val cannot be reassigned

  Line 38:         classDirectories = files(reportTasks.map { it.allClassDirs })
                   ^ Val cannot be reassigned

  Line 42:         executionData = allExecutionData
                   ^ Val cannot be reassigned

  Line 45:             executionData = allExecutionData.filter { it.exists() }
                                       ^ Type mismatch: inferred type is FileCollection! but ConfigurableFileCollection! was expected

4 errors

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.6.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 2m 23s
```
I did not find any guidelines on how to build the project. Sorry about filing this issue if the problem is on my side.

### LINKED ISSUE #97: Upgrade to Gradle 6.x
We still use Gradle 5.0, which is from 2018, so we should upgrade to a more recent version. There are some incompatibilities that need to be fixed. (See #95)

radl97:
Resolves #97 

Now it seems to be working.

First problem is related to some deprecated and removed setter. This was an easy fix when finding out the actual issue with workaround: https://github.com/gradle/kotlin-dsl-samples/issues/1265

Then gradle wrapper was updated (to 6.6.1, as it is stable according to ArchLinux folks I guess...).

Shadow also has a newer version because of the gradle upgrade, so it too was updated. *I guess it wasn't but it's very strange, because shadow did not work the same way after upgrading gradle (even though the version is fixed in buildSrc/gradle.properties)


radl97:
Hmm... There was one weird error, which might be because of a mistake I made fixing the original issue :D 

`mv: cannot stat 'subprojects/cfa-cli/build/libs/theta-cfa-cli-*-all.jar': No such file or directory`

The problem is that `-all` suffix is not generated. This is a workaround: a forced downgrade of the plugin shadow which creates the fat-jars. (we were already using an older version forced by old gradle (5.x)...)

Here are some (outdated) docs about how to override this: https://imperceptiblethoughts.com/shadow/configuration/#configuring-output-name ... However I don't know Gradle enough to be able to actually fix the issue :/ (where to put this configuration?)

hajduakos:
I'll check. We can also upgrade/replace the fat jar generating plugin with some alternate solution, as long as it can produce executable, stand-alone jar files.

hajduakos (approved):
Looks good. I rebased to the current master and force pushed. (Next time please pay attention to submit patches relative to the latest master.)

PR #99, opened 2020-11-27
xsts-dev

mondokm:
Added XSTS metrics and experimental automatic explicit variable selection for the PROD domain


subprojects/xsts-cli/src/main/java/hu/bme/mit/theta/xsts/cli/XstsCli.java
```diff
@@ -87,6 +87,9 @@
 	@Parameter(names = {"--header"}, description = "Print only a header (for benchmarks)", help = true)
 	boolean headerOnly = false;
 
+	@Parameter(names = "--metrics", description = "Print metrics about the CFA without running the algorithm")
```
hajduakos, line 90: Copy-paste: CFA

hajduakos:
Can `AutomaticItpRefToProd2ExplPredPrec` be already used? I do not see it as an option in the command line interface.

mondokm:
Not yet but I can add a command-line option

hajduakos:
I think it would be a good idea to add as an option. Otherwise, it's just there but it cannot be used. Don't forget to extend the docs as well.

hajduakos:
@mondokm When you have some time, you can also do #100 as part of this PR

hajduakos (changes requested):
I've checked the PR and seems good. However, the docs should also be updated:
- [ ] Describe the new `-metrics` argument at https://github.com/ftsrg/theta/blob/master/subprojects/xsts-cli/README.md#arguments
- [ ] Describe the new `PROD_AUTO` option and `--maxpredcount` at https://github.com/ftsrg/theta/blob/master/doc/CEGAR-algorithms.md
- [ ] Describe the PNML fronted at https://github.com/ftsrg/theta/blob/master/subprojects/xsts/README.md. You can also check out the STS related projects on how AIGER is mentioned there, because PNML is similar in this sense. It is probably a good idea to mark this feature as experimental.
- [ ] It might be worth also mentioning in the new possibility for `*.pnml` at the `--model` argument https://github.com/ftsrg/theta/blob/master/subprojects/xsts-cli/README.md#arguments

hajduakos (approved):

PR #101, opened 2021-02-01
Grouping subprojects into directories

szdan97:


hajduakos:
@szdan97 could you please rename `subprojects/core` to `subprojects/common`? I think that fits better.

PR #108, opened 2021-03-03
Test-generation feature for XTA

szkupienpeti:
Location-covering test generation for XTA models, developed in my BSc thesis.
* New subproject that contains test-generation related functionalities
* XtaAction.DELAY variable is now public (because we need to refer to that in other places)
* GraphvizWriter (and NodeAttributes) supports invisible nodes
* Command line arguments to turn on test generation and visualization

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -18,6 +18,7 @@ include(
         "xta/xta",
         "xta/xta-analysis",
         "xta/xta-cli",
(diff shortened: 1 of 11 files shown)


leventeBajczi:
Closing for now, as the integration effort is probably not worth it at this point. If anyone wants to take this up, feel free to reopen!

PR #110, opened 2021-04-10
RatToIntExpr: rational to int type conversion expression

leventeBajczi:
Added support for [mkReal2Int()](https://z3prover.github.io/api/html/classcom_1_1microsoft_1_1z3_1_1_context.html#a2da72b44de5fe681eaf8a80daf13ffdb) as a Theta expression. 
Note that it implements the UnaryExpr interface instead of the CastExpr interface to avoid automatic casts resulting in information loss.


hajduakos:
Good job. Could you please add a few test cases?
For example `core/expr/EvaluationTest.java` and `core/utils/ExprSimplifierTest.java` are good candidates to extend with a few test cases like, "is 4/3 to int equal 1?".

hajduakos (approved):

hajduakos:
@leventeBajczi before merging, could you please bump the version number? I think a patch increment is enough (`v2.12.1`). Then you can also make a lightweight release (no binary jars uploaded). @mondokm can help if needed.

hajduakos (approved):

PR #111, opened 2021-04-11
Adding a naive datalog engine to Theta

leventeBajczi:
Datalog is a logic programming language that is a more constrained subset of Prolog. It is primarily used to query knowledgebases where a (given or deduced) relation will always hold once added to the knowledgebase, which in turn allows for very efficient incremental deduction methods. See [this lecture note](http://pages.cs.wisc.edu/~paris/cs838-s16/lecture-notes/lecture8.pdf) for further information on both theory and implementation.

The main reason for this PR is to aid the development of an improved version of the verification algorithm seen in [this](https://tdk.bme.hu/VIK/ViewPaper/Memoriamodellekkel-parameterezheto) TDK paper, which uses a multi-core memory model to verify certain criteria on multi-threaded programs. Due to the nature of the algorithm, an efficient incremental evaluation is key for this approach to succeed, which is to be provided by Datalog through this PR.

Currently, this feature only supports naive evaluation, and is therefore suboptimal in time complexity. However, this can be used as a proof-of-concept in the projects utilizing it, and later can be optimized for speed when necessary.

## Contributions to Theta

This PR adds the following features to Theta:
1. Ability to run a Datalog program from a textual representation
This is mainly useful for concept evaluation and testing purposes. See `hu.bme.mit.theta.common.datalog.DatalogStringTest`, which provides two examples for simple Datalog programs.
2. Ability to create a Datalog program via code, then add facts and query relations incrementally
This is the main way to interface with Datalog from other algorithms, an example for which can be found in `hu.bme.mit.theta.common.datalog.DatalogJavaTest`

The implementation represents atoms as DatalogArgument instances, and relations as Tuples. For longer than 7-term relations, a new n-ary Tuple type is introduced, which only supports a single type of element, but can and hold arbitrarily large amount of them.


hajduakos:
@leventeBajczi can you please rebase to current master? That might fix failing tests (although I don't know why they are failing because the logs are not visible anymore, but I remember some Z3 bintray issues were fixed on master).

leventeBajczi:
This PR is included in the #157 PR.

PR #120, opened 2021-08-01
Check CFA variables and locations for unique names

hajduakos:
Previously we could have locations and variables in the same CFA with the same name. This was a bit fragile, because if we serialize the CFA and then parse it back, we might get weird errors (e.g., type of a variable does not match). To prevent these, now the CFA and the CfaBuilder enforces unique variable and location names. Intermediate locations (introduced as a syntactic sugar) now also have names. I bumped the minor version number as this might be a breaking change for frontends. Although if some frontend was relying on this previously that was a bad practice.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ buildscript {
 
 allprojects {
     group = "hu.bme.mit.inf.theta"
-    version = "2.16.0"
+    version = "2.17.0"
 
     apply(from = rootDir.resolve("gradle/shared-with-buildSrc/mirrors.gradle.kts"))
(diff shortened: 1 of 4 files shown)


leventeBajczi (approved):
Seems alright, we'll see if it breaks anything in the new C frontend and adapt it as necessary.

mondokm (commented):
It seems good to me, but I will let @leventeBajczi take a look as well.

hajduakos:
Thanks for the quick response. Codacy found an unused import (my mistake). I just fixed that, and then I'm gonna merge.

PR #123, opened 2021-08-02
Adding Support for Floating Point Arithmetic  

leventeBajczi:
The FloatingPoint type (see: [FloatingPoint Theory](http://smtlib.cs.uiowa.edu/theories-FloatingPoint.shtml)) adds support for the IEEE-754 Floating-Point Arithmetic standard, which defines a portable fp-arithmetic used (among others) by C/C++.

This PR contains the following:

1. A new Type: [FpType](https://github.com/ftsrg/theta/blob/fp/subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/fptype/FpType.java)
2. A new dependency: mpfr (necessary for BigFloat), see [this](https://github.com/ftsrg/theta/commit/7281a541c257d608fa6b2bca222df86202c70612) commit
3. New Exprs for handling FpTypes: see [this](https://github.com/ftsrg/theta/tree/fp/subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/fptype) package
4. Solver mappings to utilize the new type
5. Tests to verify correct implementation of all the above, see [FpTestUtils](https://github.com/ftsrg/theta/blob/fp/subprojects/common/core/src/main/java/hu/bme/mit/theta/core/utils/FpTestUtils.java)

Todos before merge:

- [x] Implement simplifications in [ExprSimplifier](https://github.com/ftsrg/theta/blob/fp/subprojects/common/core/src/main/java/hu/bme/mit/theta/core/utils/ExprSimplifier.java) @AdamZsofi 
- [x] Review existing parts
- [x] Add operators in the CFA grammar (pending until the review of existing parts is done, as to avoid inconsistent implementation)


leventeBajczi:
After a handful of tests, the following issues were uncovered:

- As FPSort uses bitvectors, interpolation is not supported. Therefore most refinement strategies produce an exception.
- Some refinement strategies produce an Out-of-Bounds exception in the ArgTracer class (such as NWT_IT_WP)
- The working NWT_IT_* strategies keep refining seemingly indefinitely, with no real progress
- The refinement strategies that do work can only handle the simplest of tasks (NWT_WP, NWT_SP, NWT_WP_LV), up to around 10 expressions - any more and they time out (I tested with a 10 minute time limit)

hajduakos (approved):
I didn't look into all the details, but from a high-level point it looks good. I also see that it has been thoroughly tested (especially evaluation). Thanks for the implementation! I only have one remark, but that can be done later in a separate PR. As far as I see, testing is done similarly to bitvectors, but that is a bit different from other types/expressions. For example, for the rest of the types, evaluation is tested in expr/EvaluationTest..java, but for Bv and Fp we have them under type/...TypeTest.java. Both approaches seem fine (i.e., splitting tests by feature (like evaluation) or splitting by type), but ideally they should be unified. Also, I do not understand at a first glance why some test utilities are in src/main instead of src/test. Maybe we can create an issue for this unification.

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/utils/FpTestUtils.java
```diff
@@ -0,0 +1,181 @@
+package hu.bme.mit.theta.core.utils;
```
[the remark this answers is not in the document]
leventeBajczi: It is used both by FpTypeTest and by Z3SolverFPTest, and therefore it is only accessible by both this way (if it was in any of the test packages, only one subproject could access to AFAIK)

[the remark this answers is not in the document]
radl97: Gradle can be convinced otherwise, if it is not the default setting.

If I am not mistaken, adding something like this line could help:
https://github.com/ftsrg/theta/blob/master/subprojects/cfa/cfa-analysis/build.gradle.kts#L10

leventeBajczi:
Thanks for the review, I'll wait for @as3810t to take a look as well and then I'll merge it.

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/utils/FpTestUtils.java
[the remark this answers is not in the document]
leventeBajczi: Thanks, I'll see what I can do about it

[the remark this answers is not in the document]
leventeBajczi: We arrived at the conclusion that it is not worth the hassle, because Gradle does not want to cooperate. So for now this class shall remain in its original place.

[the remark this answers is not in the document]
radl97: I've thought I've came across this issue some time ago, but it is a more complex one... I was mistaken: testCompile/testImplementation does not do this (it is created for importing testing utilities, such as JUnit)...

A clear solution would be to create a core-test-utils subproject, which exports the test thingies.
This would blow up easily (I mean combinatorically: for every subproject for one or two test files, a whole subproject...?)...

There could also be some Gradle magic, with which one can import the `test` from another project, but I do not know how to achieve this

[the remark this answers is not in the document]
leventeBajczi: With this I could successfully move the class out of the _main_ source-set. Thanks for the help!

subprojects/cfa/cfa/src/main/antlr/CfaDsl.g4
```diff
@@ -214,7 +223,7 @@ additiveExpr
 	;
 
 multiplicativeExpr
-	:	ops+=bvConcatExpr (opers+=(MUL | DIV | MOD | REM | BV_MUL | BV_UDIV | BV_SDIV | BV_SMOD | BV_UREM | BV_SREM) ops+=bvConcatExpr)*
+	:	ops+=bvConcatExpr (opers+=(MUL | DIV | MOD | REM | BV_MUL | BV_UDIV | BV_SDIV | BV_SMOD | BV_UREM | BV_SREM | FPREM) ops+=bvConcatExpr)*
```
as3810t, line 226: I have one question regarding this. How can you specify the rounding mode?

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/fptype/FpType.java
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	@Override
+	public String toString() {
+		return Utils.lispStringBuilder(TYPE_LABEL).add(exponent).add(significand).toString();
+	}
+
+	@Override
+	public AddExpr<FpType> Add(Iterable<? extends Expr<FpType>> ops) {
+		return FpExprs.Add(FpRoundingMode.getDefaultRoundingMode(), ops);
```
as3810t, line 93: This default rounding mode has some implications and I am not sure whether this is intentional. This function is called by the grammar mainly to create the expressions it sees. So this line of code (and the other similarly) mean that you can only create expressions of one specific rounding type through the grammar.

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/fptype/FpType.java
```diff
@@ -0,0 +1,140 @@
...
+import hu.bme.mit.theta.core.type.abstracttype.Multiplicative;
+import hu.bme.mit.theta.core.type.abstracttype.NegExpr;
+import hu.bme.mit.theta.core.type.abstracttype.NeqExpr;
+import hu.bme.mit.theta.core.type.abstracttype.Ordered;
+import hu.bme.mit.theta.core.type.abstracttype.PosExpr;
+import hu.bme.mit.theta.core.type.abstracttype.SubExpr;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+public class FpType implements Equational<FpType>, Additive<FpType>, Multiplicative<FpType>, Ordered<FpType> {
```
as3810t, line 24: I am not sure, that Additive and Multiplicative is a wise choice here. Yes, I understand that the semantics align, but actually Theta's grammar use these interfaces to create expressions. This means, that the `+` operator will create an FpAddExpr if both operands are floats. And this way, you cannot specify the rounding modes.

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/fptype/FpAddExpr.java
```diff
@@ -0,0 +1,84 @@
...
+import hu.bme.mit.theta.core.type.abstracttype.AddExpr;
+import hu.bme.mit.theta.core.utils.TypeUtils;
+
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static hu.bme.mit.theta.core.utils.TypeUtils.checkAllTypesEqual;
+
+public class FpAddExpr extends AddExpr<FpType> {
```
as3810t, line 14: I am not sure about AddExpr. See my comment in FpType.java

subprojects/cfa/cfa/src/main/java/hu/bme/mit/theta/cfa/dsl/CfaExpression.java
```diff
@@ -705,17 +757,30 @@ private BvConcatExpr createBvConcatExpr(final Expr<BvType> leftOp, final Expr<Bv
...
 						return Neg(op);
 
+					case FP_ABS:
+						return Abs((Expr<FpType>) op);
+
+					case FP_IS_NAN:
+						return IsNan((Expr<FpType>) op);
+
+					case FPROUNDTOINT:
+						return RoundToIntegral(FpRoundingMode.RNE, (Expr<FpType>) op);
```
as3810t, line 779:
```suggestion
						return RoundToIntegral(FpRoundingMode.getDefaultRoundingMode(), (Expr<FpType>) op);
```

subprojects/cfa/cfa/src/main/java/hu/bme/mit/theta/cfa/dsl/CfaExpression.java
```diff
@@ -705,17 +757,30 @@ private BvConcatExpr createBvConcatExpr(final Expr<BvType> leftOp, final Expr<Bv
...
+
+					case FPSQRT:
+						return Sqrt(FpRoundingMode.RNE, (Expr<FpType>) op);
```
as3810t, line 782:
```suggestion
						return Sqrt(FpRoundingMode.getDefaultRoundingMode(), (Expr<FpType>) op);
```

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/fptype/FpRoundingMode.java
```diff
@@ -0,0 +1,13 @@
...
+
+public enum FpRoundingMode {
+	RNE, /* Round nearest ties to even */
+	RNA, /* Round nearest ties to away */
+	RTP, /* Round toward positive */
+	RTN, /* Round toward negative */
+	RTZ; /* Round toward zero */
+
+	public static FpRoundingMode getDefaultRoundingMode() {
+		return RNE;
```
as3810t, line 11: This particular rounding mode is the default according to the C standard. Maybe it is worth mentioning somewhere in the docs or a comment here.
```suggestion
		return RNE; /* Default because it is default in the C standard */
```

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/utils/IEEEBigDecimal.java
```diff
@@ -0,0 +1,346 @@
...
+	public String toString() {
+		if (this == NaN) return "NaN";
+		else if (this == POSITIVE_INFINITY) return "+Infinity";
+		else if (this == NEGATIVE_INFINITY) return "-Infinity";
+		else {
+			assert bigDecimal != null;
+			return bigDecimal.toString();
+		}
+	}
+}
```
as3810t, line 346: I am not sure this file is used at all. As I recall, this was an early attempt at implementing the fp standard in java. Without much success I might add. I think it can be deleted.
```suggestion
```

as3810t, line 346: I am not sure this file is used at all. As I recall, this was an early attempt at implementing the fp standard in java. Without much success I might add. I think it can be deleted.
```suggestion

```

as3810t, line 346: I am not sure this file is used at all. As I recall, this was an early attempt at implementing the fp standard in java. Without much success I might add. I think it can be deleted.
```suggestion
```

as3810t, line 346: I am not sure this file is used at all. As I recall, this was an early attempt at implementing the fp standard in java. Without much success I might add. I think it can be deleted.

subprojects/common/core/src/testFixtures/java/hu/bme/mit/theta/core/utils/FpTestUtils.java
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	public static Collection<?> NaNOperations() {
+		return Arrays.asList(new Object[][]{
+				{FpAddExpr.class, Fp16NaN(), Add(RNE, List.of(Fp16NaN(), Fp16("3.4")))},
+				{FpSubExpr.class, Fp16NaN(), Sub(RNE, Fp16("5.5"), Fp16NaN())},
+				{FpPosExpr.class, Fp16NaN(), Pos(Fp16NaN())},
+				{FpNegExpr.class, Fp16NaN(), Neg(Fp16NaN())},
+				{FpMulExpr.class, Fp16NaN(), Mul(RNE, List.of(Fp16NaN(), Fp16("3.4")))},
+				{FpDivExpr.class, Fp16NaN(), Div(RNE, Fp16("7.14"), Fp16NaN())},
```
as3810t, line 143: I think we should write some tests to test the relational operations with NaN values. They are quite the pain in the *** according to the standard.

subprojects/cfa/cfa/src/main/antlr/CfaDsl.g4
```diff
@@ -227,7 +236,7 @@ bvExtendExpr
 
 unaryExpr
 	:	bitwiseNotExpr
-	|	oper=(PLUS | MINUS | BV_POS | BV_NEG) op=unaryExpr
+	|	oper=(PLUS | MINUS | BV_POS | BV_NEG | FP_ABS | FP_IS_NAN | FPROUNDTOINT | FPSQRT) op=unaryExpr
```
as3810t, line 239: There is no way to do FpToFp conversion.

as3810t (changes requested):
I reviewed the code, here are some observations that I think should be addressed.

First of all, a major one:  **!!!** I do not now the exact licence of mpfr and mpfr-java and whether we are allowed to distribute the binaries. This should be checked and if necessary, included in the docs. **!!!**

My second big observation overall is about the grammar, but pops up all around the code. Right now, the standard (`+`, `-`, ...) operators create the respecting (`FpAddExpr`, ...) expressions. The issue with this, that this way, there is no mode from the grammar to indicate the rounding mode.

This can be an intentional decision, but than later on, if rounding mode becomes needed there will be two ways ahead:

1. Modify the grammar, so these operators can take parameters, which is a rounding mode in this case.
2. Add new operators for fp operations that take a(n optional) parameter, the rounding mode.

In the first case, a lot of regression testing becomes necessary, while in the second case the question is whether to remove the current implementation breaking compatibility.

I would personally introduce new (textual) operators for the fp operations (`fpadd`, `fpsub`, ...) right now similarly to bitvectors. These could have a parameter (`fpadd[RNE]`) that specifies the rounding mode, or if the parameter is omitted, it could use the default rounding mode. This would also mean, that the helper interfaces (`AddExpr`, `Additive`, ...) should be removed from the `FpType`, and fp expression classes.

subprojects/cfa/cfa/src/main/antlr/CfaDsl.g4
leventeBajczi: Fixed in [2608512](https://github.com/ftsrg/theta/pull/123/commits/260851260a768790fd16a6f98dcd612f29bb0084)

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/fptype/FpAddExpr.java
leventeBajczi: Somewhat fixed in [2608512](https://github.com/ftsrg/theta/pull/123/commits/260851260a768790fd16a6f98dcd612f29bb0084) -- see separate discussion

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/fptype/FpType.java
leventeBajczi: Somewhat fixed in [2608512](https://github.com/ftsrg/theta/pull/123/commits/260851260a768790fd16a6f98dcd612f29bb0084) -- see separate discussion

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/fptype/FpType.java
leventeBajczi: Somewhat fixed in [2608512](https://github.com/ftsrg/theta/pull/123/commits/260851260a768790fd16a6f98dcd612f29bb0084) -- see separate discussion

subprojects/cfa/cfa/src/main/antlr/CfaDsl.g4
leventeBajczi: Fixed in [2608512](https://github.com/ftsrg/theta/pull/123/commits/260851260a768790fd16a6f98dcd612f29bb0084)

subprojects/cfa/cfa/src/main/java/hu/bme/mit/theta/cfa/dsl/CfaExpression.java
leventeBajczi: Fixed in [19a3e6b](https://github.com/ftsrg/theta/pull/123/commits/19a3e6b5a7cf198aa7c57c6ee2fd3aec2a75a257)

subprojects/cfa/cfa/src/main/java/hu/bme/mit/theta/cfa/dsl/CfaExpression.java
leventeBajczi: Fixed in [19a3e6b](https://github.com/ftsrg/theta/pull/123/commits/19a3e6b5a7cf198aa7c57c6ee2fd3aec2a75a257)

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/fptype/FpRoundingMode.java
leventeBajczi: Fixed in [0d1af43a1]()

subprojects/common/core/src/testFixtures/java/hu/bme/mit/theta/core/utils/FpTestUtils.java
leventeBajczi: Added some tests in [9733f6f](https://github.com/ftsrg/theta/pull/123/commits/9733f6fc2817b46f5078d5c11c1d6db8b2c35ceb) - are these what you had in mind?

leventeBajczi:
I've added some NaN-related tests for relational operators, modified the grammar so that rounding mode can be specified, added some CFAs to test floating point operations with, and added the concise license text of MPFR and MPFR-Java to the lib/README.md file. Could you review it again @as3810t?

subprojects/common/core/src/testFixtures/java/hu/bme/mit/theta/core/utils/FpTestUtils.java
as3810t: Yes, exactly!

as3810t (approved):

PR #128, opened 2021-08-05
Fixed simplifyGenericArrayRead

leventeBajczi:
_(Old description: Fixed simplifyGenericArrayRead to allow non-LitExpr expressions to be in an ArrayLitExpr.)_

simplifyGenericArrayRead produced an exception when a non-literal element was added to an ArrayLitExpr. This is undesirable, as neither the grammar nor the ArrayLitExpr constrained its creation with non-literal values.

This PR fixes this problem the following way:
* ArrayLitExpr no longer accepts non-literal values (for compatibility and ease-of-use any expression is accepted, but if it cannot be simplified to a LitExpr, an exception is raised)
* A new expression, ArrayInitExpr is added to handle non-literal array 'literals'. This is a generic MultiaryExpr, as this cannot be considered a literal.


subprojects/common/core/src/main/java/hu/bme/mit/theta/core/utils/ExprSimplifier.java
```diff
@@ -412,7 +414,13 @@ private ExprSimplifier() {
 		Expr<ArrayType<IT, ET>> arr = simplify(expr.getArray(), val);
 		Expr<IT> index = simplify(expr.getIndex(), val);
 		if (arr instanceof LitExpr<?> && index instanceof LitExpr<?>) {
```
hajduakos, line 416: I would add a comment here why the condition `index instanceof LitExpr<?>` is required. If I'm not mistaken it is required because for literals `equals` is precise. However, if we would have `x+1` as an index and we don't know `x`, then we cannot really decide whether to return some element with a key, or a default element. Like if your array has `1 => 1, 2 => 2, else => 3` then `arr[x+1]` can basically correspond to any element, including the `else` as well.

hajduakos (approved):

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/utils/ExprSimplifier.java
hajduakos: I think a comment like the following would do:

> The index is required to be a literal so that we can use 'equals' to compare it against existing keys in the array

hajduakos:
This might be a dummy question, but is there any example where array literals have non-literal elements? I mean the syntax allows, that's true, but is there any practical case where for example the solver is returning such an array? I assume yes, because otherwise you wouldn't open this issue. I'm just curious to see the actual example.

leventeBajczi:
Yes, I came across this issue while developing the C frontend. Consider the following program:
```
 int main() {
    int x;
    int a[3] = {2, x+1, 4};
    if(a[1] > 2) reach_error();
 }
```
Here, we cannot assume anything about the 2nd element - so we can handle it as an AddExpr with `x` and `1` as operands. This however breaks the current simplification approach.

leventeBajczi:
Some notes:

- ArrayInitExpr became a MultiaryExpr, but at the cost of having a generic `Type` OpType. The ops are used in a non-standard way, for more info, see the helper comment at the top of the class.
- The grammar did not change, as arbitrary expressions were already allowed to be specified. The CfaExpression visitor did change though, as firstly all array 'literals' get mapped onto an ArrayInitExpr, which is then simplified, effectively yielding ArrayLitExprs when possible.
- Testing was not too thorough, but these changes did not seem to break anything. Some evaluation, z3 and CFA tests were added. Is there anything else worth testing?

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/arraytype/ArrayInitExpr.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * ArrayInitExpr is a way to specify arbitrary array 'literals' that may contain non-literal elements as well.
+ * Note that while this class is a descendant of MultiaryExpr, it is used in a non-standard way: ops is only used as a
+ *     generic Type type, and they are solely used for inter-object interactions, intra-class the `elems` and `elseElem`
```
hajduakos, line 23: Can you please fix the formatting here?

subprojects/common/core/src/test/java/hu/bme/mit/theta/core/utils/ExprSimplifierTest.java
```diff
@@ -744,6 +751,21 @@ public void testArrayWrite() {
 		assertEquals(Int(100), Read(newArr, Int(182)).eval(ImmutableValuation.empty()));
 	}
 
+	@Test
```
hajduakos, line 754: As far as I see, this test checks whether we can read(init(...)) can be simplified or not. Can you also include a test when the array init expression only contains literals and calling simplify makes it an arraylitexpr?

subprojects/common/solver-z3/src/test/java/hu/bme/mit/theta/solver/z3/Z3SolverTest.java
```diff
@@ -152,6 +155,37 @@ public void testArray() {
 		assertEquals(Int(2), Read(valLit, Int(1)).eval(ImmutableValuation.empty()));
 	}
 
+	@Test
+	public void testArrayInit() {
+        final ConstDecl<ArrayType<IntType, IntType>> arr = Const("arr", Array(Int(), Int()));
+		var elems = new ArrayList<Tuple2<Expr<IntType>, Expr<IntType>>>();
```
hajduakos, line 161: Formatting seems to be a bit inconsistent.

hajduakos (changes requested):
Looks good in general. The solution for making ArrayInitExpr multiary is indeed a bit weird, but I don't have a better idea for now. I think it's fine as long as it works and it's explained. We can change the implementation later when we have a better solution.

I requested some minor changes.

subprojects/common/solver-z3/src/test/java/hu/bme/mit/theta/solver/z3/Z3SolverTest.java
leventeBajczi: I don't know how I managed that, because in IntelliJ it looked fine, but I fixed it now.

subprojects/common/core/src/test/java/hu/bme/mit/theta/core/utils/ExprSimplifierTest.java
leventeBajczi: I've added some LOC testing this behavior, but I also kept the existing tests - is it OK like this? https://github.com/ftsrg/theta/pull/128/commits/c81108dccfd8e3799c058554e36360f6e7fabfaa

hajduakos (approved):
Looks good!

hajduakos:
Don't forget to bump the minor version number.

PR #129, opened 2021-08-06
Syntactic sugar for array writes in XSTS

mondokm:
This pull request introduces the syntactic sugar `array[index]:=value` which translates to `array:=array[index<-value]`


hajduakos:
@mondokm I see that you bumped the version number, but there is no release (yet). Is that in progress?

mondokm:
Yes, it's in progress, but I bumped into a different issue regarding building on windows, which I wanted to take care of first

PR #130, opened 2021-08-06
Add Windows build to CI, disable float tests on Windows.

hajduakos:
This PR adds a basic Windows build to CI (GitHub actions). Currently it's failing due to a missing library for floats:
```
java.lang.UnsatisfiedLinkError at Z3SolverFPTest.java:45
```

But if we fix that and rebase this PR, I think it should work. But let's wait for that fix first.


hajduakos:
See #131

hajduakos:
Update: we skip float tests on Windows for now, until we can somehow port the required library to Windows.

PR #132, opened 2021-08-07
Fix win64 libraries for floating point library

leventeBajczi:
Currently the MPFR_JAVA library only works on Linux, as there are no win64 libraries available. This PR adds support for this library on the win64 platform.


hajduakos (changes requested):
Can you please rebase to master, and remove the remarks from the readme files that suggest that floats are not working on Windows?

leventeBajczi:
I hope I did everything we've talked about, but please check if everything is OK

hajduakos (approved):
Looks good! But I'd update the minor version number (2.20.0 instead of 2.19.1) as this is actually an improvement visible outside.

PR #133, opened 2021-08-09
Added notes about fp support

AdamZsofi:
FpType was added in #123 but the refinement strategies readme was not updated. This PR fixes this shortcoming.


hajduakos:
It is a good summary on the main limitations. But I think number two (out of bounds exception) is actually a bug that should be instead filed as an issue. I think that is something that can be fixed on a short term.

hajduakos (approved):

AdamZsofi:
I could not reproduce the `ArgTracer` Out-of-bounds exception issue again, something changed since then (accidental fix, hopefully). 
If it pops up again somewhere, I'll make sure to open an issue.

PR #137, opened 2021-08-30
improved ArgTrace's creation speed

asztrikx:
Instead of inserting at first index reverse at the end.

Also did some benchmark on "selected models", there's no consistent, noticeable difference.


hajduakos:
Looks good. Can you please bump the patch version number?

hajduakos (changes requested):
Can you please bump the patch version number?

asztrikx:
done, thanks for warning

hajduakos (approved):

PR #109, opened 2021-03-30
SMT-LIB solver integration

as3810t:
This pull request adds a new solver implementation next to the already existing Z3 solver. This new implementation uses the [SMT-LIB](http://smtlib.cs.uiowa.edu/) standard to communicate with the SMT solvers, so the underlying solver becomes changeable as long as it supports SMT-LIB.

This is an alpha feature, with known shortcomings. These shortcomings will need to be addressed in the future (see #71).

### The main features of this pull request:

- SMT-LIB solver integration
  - Support for booleans, integers, rationals, arrays, bitvectors, floating-points, quantifiers and unary uninterpreted functions
  - Support for unsat cores
  - Interpolation support (where applicable)
    - Binary, sequence, tree interpolation
- CLI tool of solver installer scripts for easy solver management, with support for
  - Boolector
  - CVC4
  - MathSAT
  - Princess
  - SMTInterpol
  - Yices2
  - Z3
- Integration with the CFA-CLI tool
  - Allows specifying different solvers for abstraction and refinement phase

Issue #71, opened 2020-08-29
Generic, SMT-LIB-based backend for solvers


hajduakos:
Currently we have a generic solver interface, realized by a wrapper for Z3 that uses the Java API of Z3. While using such APIs is efficient, it requires a significant effort to integrate a new solver. It would be nice to have a generic implementation, which serializes the expressions as SMT-LIB string (possibly to an in-memory file for efficiency) and passes it to an arbitrary binary that can consume the format. This would make it much more easier to integrate new solvers.

hajduakos:
@as3810t I assigned you because you already started working on it on the `smtlib-solver` branch

as3810t:
This feature is rather huge, so it will be implemented over multiple pull requests to the master, to gather feedback between steps.

### Road to SMT-LIB solvers

- [x] Initial implementation ( #109 )
  - [x] Works on Linux
  - [x] Works with some, SMT-LIB compliant solvers (CVC4, recent Z3)
  - [x] Works with other SMT-LIB compliant solvers (older Z3, MathSAT5, Princess, SMTInterpol, Yices2, Boolector)
  - [x] Works with interpolation supported on some solvers (older Z3, MathSAT5, Princess, SMTInterpol)
  - [x] Works in the CFA cli frontend
- [ ] Integration with the rest of the CLI tools
  - [ ] STS
  - [ ] XSTS
  - [ ] XTA
- [ ] Works on Windows (maybe MacOS) with some solvers (not all support Windows / MacOS)
- [ ] Testing, testing, testing
  - [ ] First of all, there should be a test suite targeting different solvers and versions and testing their supported features for regression testing. The "soul" of the integration is an ANTLR grammar, which can easily break. The problem is that the number of supported features are huge, plus there is a huge amount of solver versions, across which the supported features change. The test suite should be designed in a way, that most tests are reusable.
  - [ ] Second, a regression testing should be performed to check, that the algorithms indeed work, and there are no glaring mistakes in the implementation.
- [ ] Code cleanup

### Nice to have features

This list contains features that would be nice to have, but require big, architecture changing decisions to implement, so they are out-of-scope.

- There is no theory detection, so the solving process is not always the best.
- If using unsupported features, the failure is detected by the solver. It would be better, if Theta could catch these before sending the unsupported command down to the solver.

leventeBajczi:
I think this was solved with #143, closing.

PR #143, opened 2021-09-28
SMT-LIB solver integration

as3810t:
This pull request adds a new solver implementation next to the already existing Z3 solver. This new implementation uses the [SMT-LIB](http://smtlib.cs.uiowa.edu/) standard to communicate with the SMT solvers, so the underlying solver becomes changeable as long as it supports SMT-LIB.

This is an alpha feature, with known shortcomings. These shortcomings will need to be addressed in the future (see #71).

### The main features of this pull request:

- SMT-LIB solver integration
  - Support for booleans, integers, rationals, arrays, bitvectors, floating-points, quantifiers and unary uninterpreted functions
  - Support for unsat cores
  - Interpolation support (where applicable)
    - Binary, sequence, tree interpolation
- CLI tool of solver installer scripts for easy solver management, with support for
  - Boolector
  - CVC4
  - MathSAT
  - Princess
  - SMTInterpol
  - Yices2
  - Z3
- Integration with the CFA-CLI tool
  - Allows specifying different solvers for abstraction and refinement phase

Supersedes #109.


as3810t:
@hajduakos This version introduces major changes in alpha version. Which would you prefer: to bump only the minor version (2.24.0) or the major as well (3.0.0).

as3810t:
Based on our discussion with @hajduakos, this change is big enough to warrant `3.0.0` as it introduces new modules and breaking changes in the Solver API.

PR #146, opened 2021-11-15
Update jacoco workflow

radl97:
Bash tries to understand `**/jacoco*.xml`, but find is the one who will parse the asterisks.

It leads to a ambigious error message `**/jacoco*.xml not found`


hajduakos:
Thanks, merged. Since it's just CI change (no change to the actual code) I don't think a version number bump is needed.

PR #149, opened 2021-11-29
Fixed ArrayInitExpr withOps bug

mondokm:
Fixed a bug in ArrayInitExprs that caused the withOps method to try to cast all operands to the same type.


hajduakos (approved):
So the main issue was that with array init exprs, the index and the element types can be different, right?

mondokm:
Yes, and because of this Expr::getType can't be used, the types have to be determined based on positions.

Issue #148, opened 2021-11-29
[CFA] Floating point type syntax shadows bitvector `extract` expression


leventeBajczi:
Currently, both the type declaration of floating point variables and the `extract` expression of bitvectors use the `[FROM:TO]` notion, but for different purposes, with separate token types. This makes `extract` operations unparsable. 
Currently, some of the tests regarding CFA-s are disabled for this purpose, which must have been a quick fix, but somehow stayed in the current code on master. To my knowledge, this problem only appears in the CFA grammar [CfaDsl.g4](https://github.com/ftsrg/theta/blob/master/subprojects/cfa/cfa/src/main/antlr/CfaDsl.g4).

leventeBajczi:
@as3810t if you have some time, could you take a look? What should be the solution (what syntax could be better, and for which expression)?

as3810t:
Personally, I chose the `[FROM:TO]` syntax for the bitvector extract expression because of Python: in a way, it extracts a list of bits from a list of bits. Also, this syntax is integrated with and used by PLCverif, so I would not change it unless absolutely necessary.

As for the floating point declaration, how about `[SIG, EXP]`? Conceptually, these declarations should have two parameters, so it is given in a two-element list. On the other hand, bitvectors require a single parameter, so they are given a one-element list (`[SIZE]`).

PR #150, opened 2021-11-30
FPType grammar fix (#148)

leventeBajczi:
This PR aims to fix the problems outlined in #148. Namely:
* Distinguish `FP_TYPE_DECL` from `bvExtractAccess` by replacing the colon with a comma in `FP_TYPE_DECL`
* Add missing test cases to DSL and analysis test managers

A minor nuisance with the new grammar is the the following is accepted: `[1,2]`, while `[1, 2]` is not. Is this a compromise we are okay with, or should it be fixed? This would require a deeper restructuring of the grammar.


as3810t (approved):
I am comfortable with this restriction for now. Later on, after a bit of restructuring, `[1, 2]` can be supported as well, which would not break existing support.

PR #158, opened 2022-02-21
Added Github Actions task for copyright checking 

leventeBajczi:
This PR adds a new GH Action which checks pull requests for the presence of a copyright header in each java and kotlin file. Moving forward, this should be mandatory for all PRs before merge.

### DIFF
--- a/.github/workflows/check-copyright.yml
+++ b/.github/workflows/check-copyright.yml
@@ -1,0 +1,18 @@
+name: Copyright check
+
+on: [pull_request]
+
+jobs:
+  check-copyright:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04]
+    runs-on: ${{ matrix.os }}
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/check-copyright.yml
```diff
@@ -0,0 +1,18 @@
+name: Copyright check
+
+on: [pull_request]
```
as3810t, line 3: Do we need to execute it on every PR, or PRs to the master is enough? I would also add the `workflow_dispatch` trigger so that this task could be executed manually if needed.

```suggestion
on:
  pull_request:
    branches: [ master ]
  workflow_dispatch:
```

as3810t (approved):
I think that it is enough to execute this task only on pull requests to the master branch. But I am not against executing it more frequently, if you think that it is appropriate.

leventeBajczi:
I think it is enough if we execute it in PRs to master only. I have committed your suggested changes.

PR #163, opened 2022-03-30
Disabled windows-2016 check, added windows-2022

leventeBajczi:
Can we safely discard the windows-2016 check? It seems to get cancelled on its own for the last few commits.


as3810t (approved):
Seems OK to me. I think this can be merged.

PR #169, opened 2022-05-23
DPOR (Dynamic Partial Order Reduction)

csanadtelbisz:
The result of my BSc project laboratory: dynamic partial order reduction algorithm in XcfaLts (interleavings) to reduce the size of the abstract state space.
See [documentation](https://github.com/ftsrg/theta/files/8755516/bsc-6-project-laboratory-summary.pdf).

### DIFF
--- a/subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/XcfaCli.java
+++ b/subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/XcfaCli.java
@@ -56,6 +56,7 @@ import hu.bme.mit.theta.xcfa.analysis.common.XcfaConfig;
 import hu.bme.mit.theta.xcfa.analysis.common.XcfaConfigBuilder;
 import hu.bme.mit.theta.xcfa.analysis.common.XcfaPrec;
 import hu.bme.mit.theta.xcfa.analysis.common.XcfaState;
+import hu.bme.mit.theta.xcfa.analysis.impl.interleavings.XcfaLts;
 import hu.bme.mit.theta.xcfa.analysis.portfolio.ComplexPortfolio;
 import hu.bme.mit.theta.xcfa.analysis.portfolio.Portfolio;
 import hu.bme.mit.theta.xcfa.analysis.portfolio.SequentialPortfolio;
@@ -204,6 +205,12 @@ public class XcfaCli {
 	@Parameter(names = "--algorithm", description = "Algorithm to use when solving multithreaded programs")
 	XcfaConfigBuilder.Algorithm algorithm = XcfaConfigBuilder.Algorithm.SINGLETHREAD;
 
+	@Parameter(names = "--lbe", description = "Large-block encoding level")
+	SimpleLbePass.LBELevel lbeLevel = SimpleLbePass.LBELevel.NO_LBE;
+
+	@Parameter(names = "--por", description = "Partial order reduction enabling for multithreaded programs")
+	XcfaLts.POR_MODE porMode = XcfaLts.POR_MODE.POR_OFF;
+
 	//////////// SMTLib options ////////////
 
 	@Parameter(names = "--smt-home", description = "The path of the solver registry")
@@ -220,9 +227,6 @@ public class XcfaCli {
 
 	@Parameter(names = "--validate-abstraction-solver", description = "Activates a wrapper, which validates the assertions in the solver in each (SAT) check. Filters some solver issues.")
 	boolean validateAbstractionSolver = false;
-
-	@Parameter(names = "--lbe", description = "Large-block encoding level")
-	SimpleLbePass.LBELevel lbeLevel = SimpleLbePass.LBELevel.NO_LBE;
 
 	//////////// CONFIGURATION OPTIONS END ////////////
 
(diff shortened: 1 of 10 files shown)


subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/impl/interleavings/XcfaPorLts.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class XcfaPorLts extends PorLts<XcfaState<?>, XcfaAction, XcfaEdge> {
+
+	private final XCFA xcfa;
+
+	private final XcfaLts simpleXcfaLts;
+
+	private final Random random = new Random();
```
leventeBajczi, line 22: If not too complicated, the way to choose a transition should be configurable -- e.g. randomly (as it is now, if I understand correctly), round-robin style or one-by-one (eagerly explore a thread, then continue with another).

subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/XcfaCli.java
```diff
@@ -204,6 +206,12 @@ public class XcfaCli {
 	@Parameter(names = "--algorithm", description = "Algorithm to use when solving multithreaded programs")
 	XcfaConfigBuilder.Algorithm algorithm = XcfaConfigBuilder.Algorithm.SINGLETHREAD;
 
+	@Parameter(names = "--por", description = "Enable/disable partial order reduction for interleaving based multithreaded programs")
```
leventeBajczi, line 209: I think this should be a new algorithm (see --algorithm) instead

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/utils/FrontendXcfaBuilder.java
```diff
@@ -790,14 +790,14 @@ public XcfaLocation visit(CWhile statement, ParamPack param) {
 		builder.addLoc(outerBeforeGuard);
 		propagateMetadata(statement, outerBeforeGuard);
 		for (int i = 0; i < (UNROLL_COUNT == 0 ? 1 : UNROLL_COUNT); ++i) {
-			if (((CCompound) body).getcStatementList().size() == 0) {
+			/*if (((CCompound) body).getcStatementList().size() == 0) {
```
leventeBajczi, line 793: Thanks for the fix, please delete the commented lines, we have no need for them in the future.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/XcfaPassManager.java
```diff
@@ -46,6 +46,7 @@
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
```
leventeBajczi, line 49: Is this necessary? This is the only change in this file.

leventeBajczi (changes requested):
Thanks for the PR, it shows great work!

My only bigger request (alongside the smaller ones in the review) is the following: POR should exist side-by-side with the naive interleavings-based algorithm, and therefore its addition warrants a brand new algorithm type to be introduced. Therefore, the old `interleavings` folder in xcfa-analysis/impl should be brought back with a new name (such as `naive-interleavings`, or something similar) and the new one should be `por-interleavings`. Of course, this also brings with itself the need for a revised configuration step, both in the CLI and the ConfigBuilder. If there are problems with any of those, please let me know and I'll be happy to help. Thanks!

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/utils/FrontendXcfaBuilder.java
csanadtelbisz: Commented lines removed.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/XcfaPassManager.java
csanadtelbisz: Removed. (It's a result of a debugging session where the XCFA was printed out. Sets were needed them; then I forgot to remove the import.)

subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/XcfaCli.java
csanadtelbisz: Done, see [XcfaConfigBuilder.Algorithm.INTERLEAVINGS_POR](https://github.com/csanadtelbisz/theta/blob/8a2a3305c0a88c95a622330b43e8acd4e26f5062/subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/common/XcfaConfigBuilder.java#L164). ComplexPortfolio has also been updated.
POR now works with the ```--algorithm INTERLEAVINGS_POR``` (even when complex portfolio is used).

leventeBajczi (approved):
Nice work! 
Please add the usual license header to new files, bump the version number (you can do a 4.1.0) then you can go ahead and merge this PR.

PR #170, opened 2022-06-18
Build on Ubuntu 22.04 (beta)

hajduakos:
A beta version of Ubuntu 22.04 environment is available. Are there any objections on building there as well?

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


leventeBajczi:
I think it's a good idea to include a check for that as well, but I'm not sure if #165 will change related checks in any way - @as3810t, what do you think?

as3810t:
I agree, #165 should include this change. I also think, that Ubuntu 18.04 should be abandoned (unless the BME cloud machines require it).

leventeBajczi:
Will fix together with other CI issues, closing for now.

PR #175, opened 2022-09-19
Adding no refinement progress detection to XSTS cli

AdamZsofi:
This patch adds the (already existing) functionality of stopping the analysis if there is no refinement progress to XSTS cli.
(It is merely the addition of a new flag, this feature was already in use with XCFAs.)

The check is ON by default (same as with XCFAs). It can be turned off using the `--no-stuck-check` flag.
If there is no refinement progress and the analysis is constantly rebuilding the same ARGs, it stops the analysis and throws a `NotSolvableException`.

As it is basically a runtime monitoring method, it poses a minimal overhead, but it is barely noticeable, if noticeable at all. None the less, the option to turn it off is there if it is needed, e.g. for debugging.

The monitoring is implemented a bit differently in the case of MULTI_SEQ refinement, but I could not test that with an XSTS model, as I do not know about any XSTS that gets stuck with MULTI_SEQ. (But it works with XCFAs, so I see no reason for issues with XSTS.)

@grbeni I added you as a reviewer, so you know about the new flag and exception

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -11,7 +11,7 @@ buildscript {
 
 allprojects {
     group = "hu.bme.mit.inf.theta"
-    version = "4.2.1"
+    version = "4.2.2"
 
     apply(from = rootDir.resolve("gradle/shared-with-buildSrc/mirrors.gradle.kts"))
 }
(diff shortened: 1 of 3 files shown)


grbeni (approved):
Thank you for the tagging and the information, looks good to me.

mondokm (approved):
It looks good to me, thank you!

PR #176, opened 2022-10-27
Added flags to set solver in XstsCli

leventeBajczi:
This PR adds the following flags to XstsCli:

* `--smt-home`: folder where smt solvers are located
* `--refinement-solver`: solver to use during refinement
* `--abstraction-solver`: solver to use during abstraction.

For more info, see the README in the SMTLIB subproject: [README.md](https://github.com/ftsrg/theta/blob/master/subprojects/solver/solver-smtlib-cli/README.md)


leventeBajczi:
@Rorck Could you please check if Gamma still works with the new binaries (without any modification, _and_ with some of the new solvers)?

leventeBajczi:
There were no objections (and this PR only adds functionality, it does not take any prior functionality away), so I'm merging this.

PR #183, opened 2022-12-09
CHC support

s0mark:
This is the result of my BSc degree project: a CHC frontend implemented in Theta. It can be enabled with the `--chc` flag, and different transformation methods can be chosen using the `--chc-transformation FORWARD/BACKWARD` flag. See the [documentation](https://diplomaterv.vik.bme.hu/hu/Theses/Absztrakcio-alapu-technikak-CHC-problemak) for details.


subprojects/frontends/chc-frontend/src/main/java/hu/bme/mit/theta/frontend/chc/ChcUtils.java
```diff
@@ -0,0 +1,156 @@
...
+import static hu.bme.mit.theta.core.type.inttype.IntExprs.Int;
+import static hu.bme.mit.theta.core.type.rattype.RatExprs.Rat;
+
+public class ChcUtils {
+    private static GenericSmtLibSymbolTable initialSymbolTable = new GenericSmtLibSymbolTable();
+    private static GenericSmtLibSymbolTable symbolTable;
+    private static SmtLibTypeTransformer typeTransformer = new GenericSmtLibTypeTransformer(null);
+    private static SmtLibTermTransformer termTransformer = new GenericSmtLibTermTransformer(initialSymbolTable);
+    private static CharStream charStream;
+
```
as3810t, line 46: Missing private constructor.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/XcfaPassManager.java
```diff
@@ -83,7 +83,7 @@ public class XcfaPassManager {
 //				new AssignmentChainRemoval(),
 //				new NoReadVarRemovalPass(),
 //				new GlobalVarsToStoreLoad(),
-				new UnusedVarRemovalPass(),
+//				new UnusedVarRemovalPass(),
```
as3810t, line 86: This is a global modification that will also affect C source codes. My questions are:

1. Why was this modification needed?
2. If the change is needed for CHCs, can we somehow make XcfaPassManager configurable to include the pass for C codes and remove it for CHCs.

as3810t (changes requested):
All in all, I think this is a very nice work. I left some minor comments. The only major thing missing (and we do not set a good example on this matter :sweat_smile:) is to add the copyright to the new source files you created.

See an example [here](https://github.com/ftsrg/theta/blob/9531d2858f94f92ce4e55ed189d4b5400023339c/subprojects/common/core/src/main/java/hu/bme/mit/theta/core/stmt/AssignStmt.java#L1C1-L15C4). Please do update the year though.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/XcfaPassManager.java
s0mark: The modification was necessary because the pass removed variables that would've been necessary for correct verification, resulting in incorrect answers. I added methods to XcfaPassManager that can remove passes and moved the removal to CHC handling part of XcfaCli.

s0mark:
> All in all, I think this is a very nice work. I left some minor comments.

I implemented the requested changes.

> The only major thing missing (and we do not set a good example on this matter 😅) is to add the copyright to the new source files you created. See an example [here](https://github.com/ftsrg/theta/blob/9531d2858f94f92ce4e55ed189d4b5400023339c/subprojects/common/core/src/main/java/hu/bme/mit/theta/core/stmt/AssignStmt.java#L1C1-L15C4). Please do update the year though.

I added the license text to all `.java` files in the CHC frontend. I wasn't sure about what to do with the ANTLR file for CHCs, because it is only a minor extension of the SMT-LIBv2 grammar file created by [Julian Thome](https://github.com/julianthome/smtlibv2-grammar). I ended up leaving the original MIT license, please let me know if a change is necessary.

as3810t (approved):
Thank you for the changes! I think this can be merged @leventeBajczi .

leventeBajczi (approved):
I agree! I'll go ahead and merge this PR. Thanks for the work @s0mark and the review @as3810t!

PR #198, opened 2023-07-13
XCFA-refactor 2023 [draft]

leventeBajczi:
This PR merges the `xcfa-refactor` branch.

Main features of the PR:

- [x] Adds back support for the [llvm](https://llvm.org/) backend for XCFAs
- [x] Refactors the analysis methods for programs over weakly ordered architectures
- [x] Adds the refactored runtime monitor (by @AdamZsofi )
- [x] Adds support for the [ARG web debugger](https://ftsrg-edu.github.io/student-sisak-argviz/)
- [x] Adds serialization and deserialization to all Exprs and Stmts, as well as ARGs and related data structures
- [x] Adds a new type of solver (`graph-solver`) to Theta
- [x] Adds a new frontend, `litmus-cli` for litmus test inputs
- [x] Adds lots of new tests and resources for better quality assurance throughout Theta
- [x] Rewrites the XCFA type family to use Kotlin and data classes for easier use and maintainability
- [x] Rewrites the XCFA pass system to be more resilient to reconfiguration    
- [x] Expand the documentation of various subprojects of Theta

PR #202, opened 2023-09-01
Constant variable substitution, loop unroll pass + dpor code cleanup

csanadtelbisz:
The somewhat messy code used for the TSE article is cleaned up and contributed in this pull request.


subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgNode.java
leventeBajczi, line 1: Why was this change necessary? How come this used to be so complicated, and now an ID equivalence is enough?

leventeBajczi (approved):
Nice work, thanks for the cleanup! There's one minor question (see comment on the file), but otherwise, I think these changes are mergeable.

leventeBajczi:
Also, feel free to merge the PR without waiting for all the tests - this PR does not target master, so we can merge it and fix it in #198.

subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgNode.java
csanadtelbisz: Well, I have no idea why this was so complicated beforehand. I can see no reason for the ID equivalence check being an incorrect way of determining the equivalence of ArgNodes. On the other hand, the former complicated way resulted in a depth-first ARG search from the given ArgNode (cf. line 284 in the old version comparing `outEdges` and [line 33](https://github.com/ftsrg/theta/blob/2c863d7c8d0dd0b991cbeda6f1186f3b8a5fee7e/subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgEdge.java#L33C58-L33C58) comparing the targets of ArgEdges).
(We had a conversation about this issue on the 15th of June; also with @AdamZsofi)

csanadtelbisz: Based on git blame, you are the author of the original equivalence check, so you may have the reasons for implementing equivalence check this way.
https://github.com/ftsrg/theta/blob/2c863d7c8d0dd0b991cbeda6f1186f3b8a5fee7e/subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgEdge.java#L33C58-L33C58

leventeBajczi: I see, I've made these changes in e9defab98e7436b08276ff8c5049354cece3523a. The reason for them was that even if you fully re-build an ARG from a model a second time, due to mismatched IDs, the two ARG's comparison would fail. I think that with this simplified version, the problem would reappear, and we'd have some problems with ARG comparisons, e.g., in the progress check of @AdamZsofi. Of course we can mitigate it some other way, I don't think it's acceptable that a DFS search should run at every ArgNode comparison.

leventeBajczi: Let's move this topic to our next in-person meeting. I think we will need to come up with a solution that neither breaks existing functionality nor does it place such a burden on performance.

leventeBajczi: (also, @AdamZsofi: some tests would be nice, so that similar changes are reflected in the CT pipeline's success!)

csanadtelbisz: I omit this change from this PR but I have created the issue #203 about ArgNode comparison.

PR #227, opened 2023-10-09
spor as an xcfa specific algorithm - refactor

csanadtelbisz:


csanadtelbisz:
Waiting for sanity benchmark tests before merging...

Issue #209, opened 2023-09-20
Wrong result is given for the unreach-call property


leventeBajczi:
Config 1: `--strategy DIRECT --maxenum 1 --loglevel RESULT --refinement SEQ_ITP --lbe LBE_SEQ --search ERR --prunestrategy FULL --predsplit ATOMS --cex-monitor DISABLE --domain EXPL`
Problem17_label11.yml

Config 2: `--strategy DIRECT --maxenum 1 --loglevel RESULT --refinement SEQ_ITP --lbe LBE_SEQ --search ERR --prunestrategy FULL --predsplit ATOMS --cex-monitor DISABLE --domain PRED_CART`
Problem10_label30.yml, Problem11_label40.yml, Problem15_label49.yml

s0mark:
I fixed the integer division and modulo problems in #230. I ran Theta with the configurations and the tasks mentioned in the issue, it no longer gives wrong answers to them.

This issue brings up the possibility of further discrepancies between the semantics of C, Theta and the solvers: the `sdiv`/`srem` and `udiv`/`urem` operators for bitvectors might also be effected.

leventeBajczi:
Thanks! Closing.

PR #230, opened 2023-10-23
Harmonize integer div/mod semantics

s0mark:
Fix the discrepancies between the semantics of integer division and modulo in C, Z3 and Theta (see issue #209):
- ebbb6bf: Division in IntLitExpr was calculated using `BigInteger.divide`, which follows C semantics. Converted it to SMT semantics, and added missing tests for integer division to EvaluationTest.
- be73b18: Added conversion from C semantics to SMT semantics to the C-frontend.


leventeBajczi:
Thanks! I think you can go ahead with the merge as soon as the junit tests succeed. We'll fix formatting in the other PR.

PR #232, opened 2023-10-28
WIP: Choice-else branch support

arminzavada:


leventeBajczi:
@arminzavada, would this still be useful if this got merged? If so, please add some details to the description of this PR, and we should take a look, otherwise, please close.

arminzavada:
> @arminzavada, would this still be useful if this got merged? If so, please add some details to the description of this PR, and we should take a look, otherwise, please close.

@leventeBajczi The feature itself would be really useful on my side, however, I was unable to implement it without help. 

This PR is out of date, and no real value has been added yet, so the simplest thing would be to close this pr, create a new issue, and rethink the implementation strategy, if it is even possible to implement it in a performant manner.

PR #233, opened 2023-11-01
Xcfa phased procedure pass optimization + unsupported initializer

csanadtelbisz:
waiting for benchmarks to finish before merging...


leventeBajczi (approved):
Thanks for the work, seems good!

PR #238, opened 2023-11-09
ARG equals/hashCode fix

leventeBajczi:
This PR fixes the problems with the equals and hashcode implementations of ARG, ArgTrace, ArgNode and ArgEdge.

See issue #219 for reference.

**Do not merge** until benchmarks succeed.

### DIFF
--- a/subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgEdge.java
+++ b/subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgEdge.java
@@ -19,9 +19,6 @@ import hu.bme.mit.theta.analysis.Action;
 import hu.bme.mit.theta.analysis.State;
 
 public final class ArgEdge<S extends State, A extends Action> {
-    private static final int HASH_SEED = 3653;
-    private volatile int hashCode = 0;
-
     private final ArgNode<S, A> source;
     private final ArgNode<S, A> target;
     private final A action;
--- a/subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgEdge.java
+++ b/subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgEdge.java
@@ -18,35 +18,9 @@ package hu.bme.mit.theta.analysis.algorithm;
 import hu.bme.mit.theta.analysis.Action;
 import hu.bme.mit.theta.analysis.State;
 
-import java.util.Objects;
-
 public final class ArgEdge<S extends State, A extends Action> {
     private static final int HASH_SEED = 3653;
     private volatile int hashCode = 0;
-
-    @Override
-    public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
-        ArgEdge<?, ?> argEdge = (ArgEdge<?, ?>) o;
-        return Objects.equals(source.getState(), argEdge.source.getState())
-                && Objects.equals(target, argEdge.target)
-                && Objects.equals(action.toString(), argEdge.action.toString());
-    }
-
-    @Override
-    public int hashCode() {
-        int result = hashCode;
-        if (result == 0) {
-            result = HASH_SEED;
-            result = 31 * result + source.getState().hashCode();
-            result = 31 * result + target.getState().hashCode();
-            result = 31 * result + action.toString().hashCode();
-            hashCode = result;
-        }
-        return result;
-        // return Objects.hash(source.getState(), target.getState(), action.toString());
-    }
 
     private final ArgNode<S, A> source;
     private final ArgNode<S, A> target;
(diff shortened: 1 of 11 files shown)


leventeBajczi:
I've had a realization while implementing these changes: If we use an external utility class to implement the structural hashCode/equals methods, we won't be able to cache  the hashcodes, so we'll always traverse the graph (albeit only towards the root, which is still a lot better than before). Does this bother us enough that we move the hash code calculation into the classes themselves (where we can dedicate a private field to their cached values)? Or should we try and cache it in a map somewhere? What about the equals() call -- is it inevitable to traverse all ancestors for the equality check?

subprojects/xcfa/xcfa-cli/src/test/java/hu/bme/mit/theta/xcfa/cli/XcfaCliVerifyTest.kt
```diff
@@ -78,8 +77,8 @@ class XcfaCliVerifyTest {
         fun chcFiles(): Stream<Arguments> {
             return Stream.of(
                 Arguments.of("/chc/chc-LIA-Lin_000.smt2", ChcFrontend.ChcTransformation.FORWARD, "--domain PRED_CART"),
-                Arguments.of("/chc/chc-LIA-Arrays_000.smt2", ChcFrontend.ChcTransformation.BACKWARD,
-                    "--domain PRED_CART"),
+//                Arguments.of("/chc/chc-LIA-Arrays_000.smt2", ChcFrontend.ChcTransformation.BACKWARD,
```
leventeBajczi, line 80: This test becomes stuck with the default ARG equality implementation.

leventeBajczi: clarification: _stuck_ as in not progressing until running out of heap space, not as in the progress check-related _stuck_.

szdan97:
> I've had a realization while implementing these changes: If we use an external utility class to implement the structural hashCode/equals methods, we won't be able to cache the hashcodes, so we'll always traverse the graph (albeit only towards the root, which is still a lot better than before). Does this bother us enough that we move the hash code calculation into the classes themselves (where we can dedicate a private field to their cached values)? Or should we try and cache it in a map somewhere? 

Although not as fast as caching inside the node itself, caching in a map in the structural equality class itself would be the most flexible solution. I'd start with that, and if the benchmarks show that it has too much overhead compared to direct caching, then we can move it back into the node. 

> What about the equals() call -- is it inevitable to traverse all ancestors for the equality check?

If we move the state equality check before the in-edge equality check in the node equals, and the action equality before the source equality in the edge equals, then we always start with the cheaper checks and end with the recursive checks. Thereby we could stop at the earliest possible point when two nodes _are not_ equivalent. 

As for the case when two nodes _are_ equivalent, the only optimization I can currently think of is to start with checking if the nodes are reference-equivalent, return true if they are, and check structural equivalence only after that. This way comparing nodes in the same ARG will stop when they reach a common ancestor (notice that comparing a later and an earlier CEX compares in the same ARG). 

As for distinct ARGs, we unfortunately always have to check the whole path from the root with this semantics. When comparing two whole ARGs, not nodes, we could have some kind of optimization so that we do not check the common prefixes of paths multiple times, but that seems to be more complicated than what we want to do as long as it is only used in a test.

subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgEdge.java
```diff
@@ -18,36 +18,10 @@
 import hu.bme.mit.theta.analysis.Action;
 import hu.bme.mit.theta.analysis.State;
 
-import java.util.Objects;
-
 public final class ArgEdge<S extends State, A extends Action> {
     private static final int HASH_SEED = 3653;
```
csanadtelbisz, line 22: Please remove unused HASH_SEED and hashCode fields from here (and from other modified classes as well).

csanadtelbisz (changes requested):

subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgEdge.java
csanadtelbisz: Now, that I am looking at it, it seems that the states attribute of ArgTrace is also unused (it is set but not used):
[states attribute](https://github.com/ftsrg/theta/blob/780fa90aba4a2b56d74ea08342ad00b87b107347/subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/ArgTrace.java#L42C37-L42C43) (I cannot add a review comment there because this line has not changed...)

leventeBajczi: Done, thanks!

leventeBajczi:
Changes since @szdan97's review:

* Reverted deletion of caching
* Moved state and action checks before ancestor checks
* Added null checks and self-reference checks
* Deleted old HASH_SEED and hashcode fields

I'll wait for @AdamZsofi to finish debugging the cex-monitor faults. In the meantime, if any further improvements come to mind, please do comment on them here.

subprojects/xcfa/xcfa-cli/src/test/java/hu/bme/mit/theta/xcfa/cli/XcfaCliVerifyTest.kt
csanadtelbisz: I try to debug this issue: I commented my first findings in the teams chat (it is muted by default, so you may have not noticed it).

PR #254, opened 2024-03-10
Z3 update

leventeBajczi:
This PR updates the native Z3 version to 4.13.0, while maintaining backwards compatibility by keeping the 4.3.x version as `solver-z3-legacy`. All code has been updated to use the Z3LegacySolverFactory for the time being. 

Technical details:

~~The com.microsoft.z3.jar file has been moved to `lib/legacy`, and a new version has been placed in `lib/current` (we cannot use the bare `lib` folder, as that is on java.library.path, and therefore all dependencies are automatically loaded). All `libz3.{so,dll,dylib` and `libz3java.{so,dll,dylib}` have been renamed `*legacy.*`, and new versions have been compiled and placed in the `lib` folder. Legacy versions of `libz3java` have been patched to search for `libz3legacy` instead, using manual hex editing -- further testing may be required if these modifications are stable. New version of Z3 cannot interpolate via the API, so we use the SMTLIB interface manually in the dedicated solver.~~

There is a new set of files in `lib`: `com.microsoft.z3legacy.jar`, and the corresponding `libz3{,java}legacy.{so,dll,dylib}`. These are used by the solver-z3-legacy subproject. Due to renaming symbols, there is no collision with the non-legacy versions.


leventeBajczi:
I still have to update the benchexec-test task to also package the new libraries, and also disable macos-11 testing, but otherwise, I think this PR is complete. Due to the nature of the legacy solver, I think we can ignore Sonar's duplication warning.

leventeBajczi:
On further inspection, for some reason, on macOS, some tests fail. However, I'm not able to debug these problems further than disable those tests on mac -- do we want to explore this further?

.github/actions/benchexec-test/action.yml
```diff
@@ -34,15 +34,14 @@ runs:
...
     - uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
-      name: Get JAR
+      name: Get zip
       with:
-        name: ThetaJars
-        path: jar/
-    - name: Add new jar to archive
+        name: Theta_SV-COMP
+        path: release/
+    - name: Unzip
```
as3810t, line 41: Are these changes intentional? Similarly, are the changes in this file are intentional?

.github/actions/cache-build/action.yml
```diff
@@ -16,5 +16,5 @@ runs:
     - name: build gradle
       uses: gradle/gradle-build-action@40b6781dcdec2762ad36556682ac74e31030cfe2 # v2.5.1
       with:
-        arguments: ${{ inputs.arguments }}
+        arguments: ${{ inputs.arguments }} --info --stacktrace
```
as3810t, line 19: Is this intentional?

subprojects/solver/solver-z3-legacy/README.md
```diff
@@ -0,0 +1,2 @@
+This project wraps the legacy [Z3 solver](https://github.com/Z3Prover/z3) into our common interfaces for solvers (located in the [solver](../solver) project).
+Normally, only the factory class should be used from this project to instantiate a new solver and then the common interfaces should be preferred.
```
as3810t, line 2: I think this description should be updated with information on how to produce the shipped `dll` and `so` files. As far as I understand, you had to recompile the old (`4.5.0`) version of Z3 from source, so at least this should be mentioned. Also, the exact version (Git hash), any modification you had to make. Or at the very least, it should be made clear, that downloading the binary from Z3's release page will not work with Theta.

subprojects/solver/solver-z3/src/main/java/hu/bme/mit/theta/solver/z3/Z3TermTransformer.java
```diff
@@ -264,7 +264,7 @@ private Expr<?> transformFpLit(final com.microsoft.z3.Expr term) {
         BigFloat bigFloat = new BigFloat((fpTerm).getSignificand(),
                 FpUtils.getMathContext(type, FpRoundingMode.RNE)).multiply(
                 new BigFloat("2", FpUtils.getMathContext(type, FpRoundingMode.RNE)).pow(
-                        new BigFloat((fpTerm).getExponent(),
+                        new BigFloat((fpTerm).getExponent(true),
```
as3810t, line 267: Is this a bugfix?

as3810t (changes requested):
As far as I can tell, the old solver package was basically copy-pasted and changed when needed. As of now, the new solver is not integrated into the CLI frontends.

I think two things need to be addressed:
- Documentation: I think we should document how to create the legacy dll and so files. Or at least that they are customized.
- Check the `SolverManager` of both the new and legacy Z3. Right now, they both have `"Z3"` set as a name. These should be different, if we want to integrate the new Z3 into the CLIs down the line.

.github/actions/cache-build/action.yml
leventeBajczi: Yes, because before that, all errors had to be investigated manually, gradle hides a lot of detail in the logs without these flags.

subprojects/solver/solver-z3-legacy/README.md
leventeBajczi: Sure, I'll do that.

subprojects/solver/solver-z3/src/main/java/hu/bme/mit/theta/solver/z3/Z3TermTransformer.java
leventeBajczi: This changed from 4.5.0 to 4.13.0, it does not have a 0-argument version anymore. As far as I could tell, this is the version of the call that does not change its semantics -- but I couldn't find much info on this topic, so a quick cross-check would be appreciated. (This is the `biased` param of the call).

as3810t: Yes, indeed.

as3810t (approved):

PR #275, opened 2024-07-07
This PR adds support for CHC solving via dedicated CHC solvers (including Z3)

leventeBajczi:


subprojects/solver/solver-smtlib/src/main/java/hu/bme/mit/theta/solver/smtlib/solver/SmtLibSolver.java
```diff
@@ -296,4 +330,88 @@ protected final GeneralResponse parseResponse(final String response) {
...
+            throw new SmtLibSolverException(res.getReason());
+        } else if (res.isSpecific()) {
+            final GetProofResponse getModelResponse = res.asSpecific().asGetProofResponse();
+            getModelResponse.getFunDeclarations().forEach((name, def) -> {
+                final Function<String, Type> stringToType = (String it) -> switch (it.toLowerCase()) {
+                    case "int" -> Int();
+                    case "bool" -> Bool();
+                    default ->
+                            throw new UnsupportedOperationException("Currently not supporting type " + it);
+                };
```
leventeBajczi, line 349: This needs to be done better, where do we have type parsing?

leventeBajczi: done, copied the relevant part from CHC parsing.

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/functype/FuncExprs.java
```diff
@@ -46,8 +47,8 @@ public static <ParamType extends Type, ResultType extends Type> FuncAppExpr<Para
     private static <ParamType extends Type, ResultType extends Type> FuncAppExpr<ParamType, ResultType> App(
             final Expr<FuncType<ParamType, ResultType>> func, final Expr<ParamType> paramHead, final List<Expr> paramTail) {
         if (!paramTail.isEmpty()) {
-            final var newParamHead = paramTail.get(paramTail.size() - 1);
-            final var newParamTail = paramTail.subList(0, paramTail.size() - 1);
+            final var newParamHead = paramTail.get(0);
+            final var newParamTail = paramTail.subList(1, paramTail.size());
```
as3810t, line 51: Is this change intentional?

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/functype/FuncExprs.java
```diff
@@ -57,9 +58,21 @@ private static <ParamType extends Type, ResultType extends Type> FuncAppExpr<Par
     public static <ParamType extends Type, ResultType extends Type> FuncAppExpr<ParamType, ResultType> App(
             final Expr<FuncType<ParamType, ResultType>> func, final List<Expr> params) {
         checkArgument(!params.isEmpty());
-        final var paramHead = params.get(params.size() - 1);
-        final var paramTail = params.subList(0, params.size() - 1);
+        final var paramHead = params.get(0);
+        final var paramTail = params.subList(1, params.size());
```
as3810t, line 62: Is this change intentional?

subprojects/solver/solver-smtlib/src/main/java/hu/bme/mit/theta/solver/smtlib/impl/generic/GenericSmtLibTermTransformer.java
```diff
@@ -254,6 +255,11 @@ public GenericSmtLibTermTransformer(final SmtLibSymbolTable symbolTable) {
             // Array
             put("select", exprArrayReadOperator());
             put("store", exprArrayWriteOperator());
+
+            // Proof
+            put("hyper-res", expectedFunc("hyper-res"));
+            put("asserted", expectedFunc("asserted"));
+            put("mp", expectedFunc("mp"));
```
as3810t, line 262: Are these generic across SMT-LIB, or only Horn specific? If the latter, it may be "nicer" to create a `GenericHornTermTransformer` that extends `GenericSmtLibTermTransformer`

subprojects/solver/solver-smtlib/src/main/java/hu/bme/mit/theta/solver/smtlib/solver/SmtLibSolver.java
```diff
@@ -296,4 +335,101 @@ protected final GeneralResponse parseResponse(final String response) {
...
+    }
+
+    private Expr<BoolType> extractProofExpr(Expr<?> expr) {
+        checkState(expr instanceof FuncAppExpr<?, ?>, "Proof should be function application.");
+        final var nameAndArgs = extractFuncAndArgs((FuncAppExpr<?, ?>) expr);
+        final var args = nameAndArgs.get2();
+        final var lastArg = args.get(args.size() - 1);
+        checkState(lastArg instanceof FuncAppExpr<?, ?>, "Proof should be function application.");
+        return (Expr<BoolType>) lastArg;
+    }
```
as3810t, line 434: Again. Maybe extracting these to a `HornSmtLibSolver extends SmtLibSolver` similarly how interpolation is not in this class? Not necessary, but I feel that `SmtLibSolver` is doing a lot of things.

subprojects/solver/solver-z3/src/main/java/hu/bme/mit/theta/solver/z3/Z3Solver.java
```diff
@@ -233,6 +248,68 @@ public void close() {
...
+                        proofNode.addChild(visited.get(child.getId()));
+                    }
+                }
+            }
+        }
+
+
+        return rootBuilder.build();
+    }
+
```
as3810t, line 312: Same as before. Maybe `HornZ3Solver`?

as3810t (approved):
I think this is a nice addition. However, it may be beneficial to separate Horn solving from generic solvers, similarly how interpolation is separated from them.

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/functype/FuncExprs.java
leventeBajczi: Yes, unfortunately beforehand we mixed up parameters when transforming HOL two-argument predicates to multi-parameter FOL ones. Now a `(func a b)` is equivalent to a `((func a) b)` (before, it was `((func b) a)`).

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/functype/FuncExprs.java
leventeBajczi: same as above

subprojects/solver/solver-smtlib/src/main/java/hu/bme/mit/theta/solver/smtlib/impl/generic/GenericSmtLibTermTransformer.java
leventeBajczi: These are "generic" in the sense that all proofs (at least by Z3) are given using these predicates. I think because these are not Horn specific, but rather Z3 specific (and we know other solvers like to reuse Z3's conventions..), I'm inclined to keep them here. However, I do concede that right now, only CHC solving requires proof back-transformation, so I don't know if they should stay here just because in the future we might need them for other stuff.

subprojects/solver/solver-smtlib/src/main/java/hu/bme/mit/theta/solver/smtlib/solver/SmtLibSolver.java
leventeBajczi: You're right, I'll extract these parts to a separate solver.

subprojects/solver/solver-z3/src/main/java/hu/bme/mit/theta/solver/z3/Z3Solver.java
leventeBajczi: Yes, extraction is the way. I'll do this shortly.

PR #282, opened 2024-07-17
Horn checker

leventeBajczi:
This adds a new SafetyChecker, called HornChecker. Bindings for XCFA were also added.

Right now tested and works with Eldarica, Golem and Z3/SPACER. 

Some aspects of SmtLib parsing were modified/fixed/adapted.

### DIFF
--- a/subprojects/solver/solver-smtlib/src/main/java/hu/bme/mit/theta/solver/smtlib/impl/generic/GenericSmtLibTermTransformer.java
+++ b/subprojects/solver/solver-smtlib/src/main/java/hu/bme/mit/theta/solver/smtlib/impl/generic/GenericSmtLibTermTransformer.java
@@ -15,15 +15,12 @@
  */
 package hu.bme.mit.theta.solver.smtlib.impl.generic;
 
-import com.google.common.collect.BiMap;
-import com.google.common.collect.HashBiMap;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import hu.bme.mit.theta.common.QuadFunction;
 import hu.bme.mit.theta.common.TernaryOperator;
 import hu.bme.mit.theta.common.TriFunction;
 import hu.bme.mit.theta.common.Tuple2;
-import hu.bme.mit.theta.core.decl.Decl;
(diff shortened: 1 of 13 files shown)


subprojects/solver/solver-smtlib/src/main/java/hu/bme/mit/theta/solver/smtlib/impl/generic/GenericSmtLibTermTransformer.java
```diff
@@ -723,16 +722,14 @@ protected Type transformSort(final SortContext ctx) {
...
+        vars.add(paramDecls);
     }
 
     protected void popParams(final List<ParamDecl<? extends Type>> paramDecls,
-                             BiMap<ParamDecl<?>, String> vars) {
-        for (final var paramDecl : paramDecls) {
-            vars.remove(paramDecl, paramDecl.getName());
-        }
+                             Stack<ParamDecl<?>> vars) {
+        vars.pop();
```
leventeBajczi, line 732: Is it true, that we always pop _exactly_ what we pushed last time? If so, I can simplify this method.

Fyi, this is why we need this: in smtlib, rebinding `let`-params is apparently legal:
`(let ((x 1)) (= (let ((x 0)) x) x))`

as3810t: Up until this point we did not really support `let`, only in some cases. `pushParams` and `popParams` are used for function arguments, `for-all`, `exists`, etc... For this reason, I think we _do_ pops what we pushed last time.

as3810t (approved):

subprojects/solver/solver-smtlib/src/main/java/hu/bme/mit/theta/solver/smtlib/impl/generic/GenericSmtLibTermTransformer.java
leventeBajczi: Great, thanks!

PR #283, opened 2024-07-23
Build tool upgrades and other various quality of life improvements

RipplB:
See the commit list and descriptions for details


RipplB:
@leventeBajczi @szdan97 @mondokm could you please check, if you see anything suspicious that could break something i don't know of?

leventeBajczi:
Thanks for these improvements, I think these are valuable!

leventeBajczi (dismissed):

leventeBajczi:
I don't think anything not covered by the tests are important enough to check if they don't break, so I wouldn't worry about latent problems.

leventeBajczi (approved):

PR #284, opened 2024-07-24
Added bounded portfolio

AdamZsofi:
Added a bounded portfolio really similar to the "IMC_THEN_KIND" backend option that was used on SV-COMP with EmergenTheta, but it was then later removed. 
Used solvers: Z3, CVC5:1.0.8, Mathsat:5.6.10, option name: `--portfolio BOUNDED`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,7 +28,7 @@ buildscript {
 
 allprojects {
     group = "hu.bme.mit.theta"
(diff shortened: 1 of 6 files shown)


leventeBajczi (approved):
Thanks for the new portfolio!

leventeBajczi (dismissed):

leventeBajczi (dismissed):

PR #285, opened 2024-07-26
Change all reference of codestyle to new path

RipplB:
Previous PR broke formatting checks (and most probably the reformatter action). This pull request aims to fix this issue


RipplB:
As the output of the check formatting action shows, the check now runs correctly (actually checking all files, no error)

leventeBajczi (approved):
Thanks!

leventeBajczi (dismissed):
I (hopefully) solved the issue when checks would fail when running from a fork. I still don't know why sonarcloud fails to realize it's finished, but I hope that by pushing a commit _after_ assigning the `ready_to_test` will be enough.

leventeBajczi (approved):

PR #289, opened 2024-07-31
Added CHC algorithm to Xsts

leventeBajczi:
This PR adds support for the Horn solving backend for Xstss.

Right now we don't support proof/CeX transformation, but that should be trivial, if it turns out this feature is useful.

I've added some tests, and it seems like there are no mistakes on the simple examples, but a more thorough testing would definitely be necessary.

Also, if it turns out this is useful, we could also think about useful ways of representing correctness proofs, because the horn solvers give us invariants for both the "env" and "trans" steps.


leventeBajczi:
Unfortunately, horn solvers tend not to support custom datatypes and sorts, so I replaced every enumType with an int.

mondokm (approved):
Looks good. Thank you!

mondokm (approved):

PR #305, opened 2024-10-17
Frontend update 2024

leventeBajczi:
* smallestCommonType now supports void
* Fixed struct and pointer dereferencing


subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/params/ExitCodes.kt
```diff
@@ -32,6 +33,7 @@ enum class ExitCodes(val code: Int) {
     SERVER_ERROR(202),
     PORTFOLIO_ERROR(203),
 
+    UNSUPPORTED_ELEMENT(209),
```
csanadtelbisz, line 36: Another error code `UNSUPPORTED_BY_ANALYSIS` or similar could be useful as well, as unsupported language elements may differ on the analysis level.

leventeBajczi: I agree, but this falls outside the scope of this current PR, as we currently do not keep track of language elements vs. backend capabilities.

leventeBajczi:
For some reason, an update of the IntelliJ version in the formatting script reformatted a lot of untouched files as well. I think it's OK if we merge these changes together with this PR.

csanadtelbisz (approved):
Though I cannot say I completely understand the code, it looks okay. Furthermore, benchmarks (containing more tasks than the github actions tests) confirm that everything works well.

PR #306, opened 2024-10-19
`Spotless` instead of intellij formatting

leventeBajczi:
This PR replaces the automatic format- and copyright checks, which currently mostly relied on IntelliJ IDEA's `format.sh` script, with the [spotless](https://github.com/diffplug/spotless) project. 

The main reason for this change is that a simple version update (from `2023.x` to `2024.x`) modified how the code style is applied to Java and Kotlin files. I think this is unacceptable. 

Further advantages of `spotless`:
 * It is easy to reformat code offline (using `./gradlew spotlessApply`)
 * It would be easy to set up a pre-commit hook ([example](https://github.com/diffplug/spotless/issues/178#issuecomment-351638034)) to check for unformatted code, not requiring a push and CI check
 * It uses well-established formatters such as [google-java-format](https://github.com/google/google-java-format/)
 * It can use [_ratcheting_](https://github.com/diffplug/spotless/tree/main/plugin-gradle#ratchet) to allow incremental formatting, **not requiring a reformat in the entire repo** 

Some disadvantages:
  * `Spotless` currently does not support intellij code style XMLs
  * We would either need to create an [Eclipse code style file](https://github.com/diffplug/spotless/tree/main/plugin-gradle#eclipse-jdt) to use custom rules, or we would be stuck with defaults (such as google's java style) -- although this is considered an advantage by many, as using commonly used formatting guidelines can be easier on new contributors

This PR also reverts recent changes in e37b32302b2cb72893e26b52bc63d305e999a98f and 185dc7ad29920254fc689ac68ff01952a1c65c5a. These are recent formatting commits that should never happen again. This will hopefully fix merge issues before they are dealt with in e.g. #304.

The formatting rules can be found in [java-common.build.kts](https://github.com/ftsrg/theta/blob/spotless/buildSrc/src/main/kotlin/java-common.gradle.kts). I'm open to any suggestion on what formatters, with which options to enable. Please check the [plugin's README](https://github.com/diffplug/spotless/tree/main/plugin-gradle) for options.

I would expect some insights from recent contributors to Theta, such as @mondokm, @RipplB, @AdamZsofi, @szdan97, @as3810t, @csanadtelbisz, @s0mark  (Sorry if I missed anyone!)

(Note that on this PR, the checks will most likely fail. I deliberately don't want to re-edit the reverted files.)


AdamZsofi:
So, if I understand correctly, the current state of formatting stays for all files, except the ones that will be changed by later commits.
Sounds good to me, definitely better than what happened with the reverted commit and the intellij update.

It's hard to imagine the new formatting from just config files, would it be possible to see the difference on a few example files? @leventeBajczi

leventeBajczi:
@AdamZsofi: good idea, I pushed the `spotless-demo-DONOTMERGE` branch for demo purposes: [changes](https://github.com/ftsrg/theta/compare/spotless...spotless-demo-DONOTMERGE)

PR #307, opened 2024-10-20
Added thorn creation to action

leventeBajczi:
We have a new tool this year @ SV-COMP, `Thorn`, which is a portfolio of Theta's HORN-based analyses. This PR adds support for Thorn's binary in the pipeline.


AdamZsofi (approved):
Looks good. Happy to have the --svcomp flag instead using 6 other flags.

PR #311, opened 2024-10-29
LTL checking

RipplB:
Move to CegarChecker

Convert AcceptancePredicate to kotlin

Add exceptions

Move LDG to kotlin

Move refinement to kotlin

Move abstracting, visualizer, utils to kotlin


RipplB:
This PR needs the following changes for sure:

- Resolve buchi automaton creation: current implementation is a licence clash (Owl is GPL3)
- Use the new config builders in the ltl check tests

RipplB:
- XstsLtlCli, here we should also call normalize if Ldgtrace.... is BoundedUnrolling
- Ltl2Ba interface, 1 method: String -> CFA

mondokm (approved):

PR #314, opened 2024-11-05
EmergenTheta Update

leventeBajczi:
This PR updates the Emergent portfolio for EmergenTheta.

- [x] MddChecker updates (by @mondokm)
- [x] Add binding for MddChecker for XcfaCli
- [x] Add reverse BMC, KIND, IMC
- [ ] Assemble Portfolio

PR #321, opened 2024-11-11
Portfolio update

leventeBajczi:
Changes:
- [x] portfolios have been created for this year (horn25, bounded25, complex25)
- [x] in complex25, when we encounter a multithreaded program (unreach-call), we start with a quick OC check
- [x] loopunrollpass's flag has been converted into an xcfa flag (unsafeUnroll)
- [x] OC check should be parameterized well (@csanadtelbisz)
- [x] bounded25 should be updated (will be done by merging #314)


subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/portfolio/complex25.kt
```diff
@@ -0,0 +1,1137 @@
...
+      debugConfig = portfolioConfig.debugConfig,
+    )
+
+  val forceUnrolledXcfa =
+    xcfa.optimizeFurther(
+      listOf(
+        AssumeFalseRemovalPass(),
+        MutexToVarPass(),
+        AtomicReadsOneWritePass(),
+        LoopUnrollPass(2), // TODO: how much to force unroll?
```
csanadtelbisz, line 113: 2 is okay, as it is only a first step and most errors are found with this limit anyway.

csanadtelbisz (changes requested):
In portfolio files, I suggest considering imports for config enum values to improve readability (so that we could use `CONFIG_OPTION` instead of `LongEnumClassName.CONFIG_OPTION`).

PR #335, opened 2025-01-27
Optimize IMC with Dedicated Solvers for Reachability and Fixed-Point

### LINKED ISSUE #299: Suggestion to improve implementation of IMC
In the [inner loop of IMC](https://github.com/ftsrg/theta/blob/cb77d7e6bb0e51ea00d2a1dc483fdf510dd8e8b4/subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/bounded/BoundedChecker.kt#L224-L247)
Theta uses the same solver both for checking reachability and for checking fixedpoint.
Because of this, everything is popped from the solver in preparation for the next check.
This is unfortunate because the solver cannot take any advantage of the incremental nature of the checks.

I believe it would be more efficient to have two dedicated solvers.
An interpolating solver that checks reachability and a non-interpolating solver that checks fixedpoint.
The only obstacle there is that the interpolant computed by the first solver must be added to the second solver.
I don't know how difficult this is to do in Theta.
Usually, it is possible to have a single context (like a term manager) and multiple solver instances operated over the same context.

KlevisImeri:
Resolves #299

Things extra to consider:
- The Fixed-Point solver was named `imcFpSolver`. 

- Added `popAll()` method to the `SolverBase` interface. This makes working with solvers easier. The user of the solver doesn't have to keep track of how many times `push()` was called when they just want to clear the 'stack'. This needs testing.

- In `ConfigToBoundedChecker.kt`, I used the same configuration for the Fixed-Point solver as the one for the itpSolver. This may need to be separated so the Fixed-Point solver can have its own configuration.


AdamZsofi:
Thanks for the changes!
Can you maybe mark this PR as a draft for now?
I approved the CI checks - if those work, we'll benchmark and compare the results. Let's talk about that part later in person.

AdamZsofi:
Also, you can use `gradlew spotlessApply` for formatting.

AdamZsofi:
Closing for now; may be reopened if finished in the future.

PR #337, opened 2025-02-13
Added monolthic l2s conversion

mondokz:


leventeBajczi:
Thank you @mondokz for the new transformation, I've fixed a small oversight and added support code to wire it in to xcfa-cli.

PR #354, opened 2025-03-24
OC checker improvements

csanadtelbisz:
Mainly a refactoring of OC checker analyses. Some fixes are also added as well as experimental support for certain weak memory models (WSC, TSO, PSO).


leventeBajczi (approved):
Thanks for these changes!

I think if the tests go through we can merge (I fixed them on `master`)

PR #360, opened 2025-04-02
Fixed bug related to enum type inference in XSTS

mondokm:


AdamZsofi (approved):
This is a small patch, but has quite the impact - before this, analysis on Gamma statecharts with the history pseudostate were broken, and could not be parsed. This PR fixed this.
Thanks for the help! @mondokm @RipplB

PR #375, opened 2025-06-09
Javasmt update

leventeBajczi:
Bumping javasmt version, which also enables us to drop the local `javasmt.jar`.


.github/actions/test-action/action.yml
```diff
@@ -9,13 +9,14 @@ runs:
...
-        for i in subprojects/*/*; do cp *dylib $i/; done
-        for i in subprojects/*/*; do cp *jnilib $i/; done
+        # cp lib/*dylib ./
+        # cp lib/*jnilib ./
+        # cp libmpfr.dylib libmpfr.4.dylib
+        # cp libmpfr_java-1.4.jnilib libmpfr_java.jnilib
+        # sudo ln -sf $PWD/libmpfr.4.dylib /usr/local/lib/
+        # for i in subprojects/*/*; do cp *dylib $i/; done
+        # for i in subprojects/*/*; do cp *jnilib $i/; done
+        # export DYLD_LIBRARY_PATH=$PWD/lib
```
as3810t: Do we still need this?

.github/workflows/mac-build-test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-13]
+        os: [macos-15]
```
as3810t: I think this should be a separate PR.

buildSrc/src/main/kotlin/Deps.kt
```diff
@@ -25,14 +38,27 @@ object Deps {
         val runtime = "org.antlr:antlr4-runtime:${Versions.antlr}"
     }
 
-    val z3 = "lib/com.microsoft.z3.jar"
-    val z3legacy = "lib/com.microsoft.z3legacy.jar"
+    val mpfr_java = when(getOs()) {
+        "windows" -> listOf("lib/mpfr_java-1.0.jar", "lib/mpfr_java-1.0-windows64.jar")
```
as3810t: Do we know that MPFR 1.0 and 1.4 are compatible on their binary interface? Can the newer version be built on Windows?

subprojects/common/core/src/testFixtures/java/hu/bme/mit/theta/core/utils/FpTestUtils.java
```diff
@@ -269,6 +244,31 @@ private static Collection<?> LinuxSpecificTests() {
         return Arrays.asList(
                 new Object[][] {
                     {FpRoundToIntegralExpr.class, Fp16("2.0"), RoundToIntegral(RNE, Fp16("1.5"))},
+                    {
+                        FpToBvExpr.class,
+                        BvUtils.bigIntegerToUnsignedBvLitExpr(BigInteger.TEN, 16),
+                        ToBv(RTZ, Fp16("10.9"), 16, false)
+                    },
+                    // {FpToBvExpr.class, BvUtils.bigIntegerToUnsignedBvLitExpr(BigInteger.TEN, 3),
+                    //         ToBv(RTZ, Fp16("10.9"), 3, false)},
```
as3810t: Dead code. Include rational

subprojects/common/core/src/testFixtures/java/hu/bme/mit/theta/core/utils/FpTestUtils.java
```diff
@@ -269,6 +244,31 @@ private static Collection<?> LinuxSpecificTests() {
...
+                    {
+                        FpToBvExpr.class,
+                        BvUtils.bigIntegerToSignedBvLitExpr(BigInteger.TEN, 16),
+                        ToBv(RTZ, Fp16("10.9"), 16, true)
+                    },
+                    // {FpToBvExpr.class, BvUtils.bigIntegerToSignedBvLitExpr(BigInteger.TEN, 4),
+                    //         ToBv(RTZ, Fp16("10.9"), 4, true)},
```
as3810t: Dead code. Include rational

as3810t (changes requested):
I have reviewed the changes, my findings are in the comments.

Most of them are small nitpicks and can be ignored. The only one that needs resolution is the reflection one. This method is called regularly (owing to expression simplification) so we really should do everything to avoid using reflection here due to performance considerations.

.github/actions/test-action/action.yml
leventeBajczi: Nope, good catch.

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/utils/FpUtils.java
```diff
@@ -123,4 +124,16 @@ public static FpLitExpr fromString(final String value, final FpType type) {
...
+        RoundingMode r = FpUtils.getMathContextRoundingMode(roundingMode);
+        try {
+            Method method = value.getClass().getMethod("toBigInteger", RoundingMode.class);
+            return (BigInteger) method.invoke(value, r);
+        } catch (NoSuchMethodException e) {
+            throw new RuntimeException("No suitable toBigInteger method found on BigFloat.", e);
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to invoke toBigInteger method on BigFloat.", e);
+        }
+    }
```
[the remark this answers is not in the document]
leventeBajczi: I did something ill-advised: we're now using the 1.4 java library with the 1.0 native library on Windows. Oh well :sweat:

PR #376, opened 2025-06-24
Serializable class hierarchy for statements, expressions and types

csanadtelbisz:
Making the statements, expressions and types serializable. For this, the relevant classes were converted into Kotlin.

Important changes:
- The above classes are Kotlin data classes now that are serializable. Serialization is not automatic (it can only be automatic with sealed classes but we have expressions defined in different modules so this is not an option). A KSP task is introduced that automatically generates a serialization module containing serializers for the classes in the core module. These serialization modules (e.g., `hu.bme.mit.theta.core.type.generated.expr.exprSerializerModule`) can be used for serialization. Note that the serialization module has to be extended whenever a new expression/type is declared in a different (gradle project) module. See the core serialization module definition in [SerializerModule.kt](https://github.com/ftsrg/theta/blob/daee0758a33fa13df29e5bb34050cf9d081e6417/subprojects/common/core/src/main/kotlin/hu/bme/mit/theta/core/serialization/SerializerModule.kt) that can be included in extended module definitions.
- Hash codes and equals function implementations are removed from the above classes: they are replaced by the auto-generated ones for Kotlin data classes. The equals (and/or hashcode implementations) are preserved when it does not trivially compares the attributes (e.g., signedness is not considered in some classes). Declarations such as VarDecl or ParamDecl were only considered equal when we had the same references: since this is not representable in a serialized form of the instances, a new `id` attribute is added to `Decl` for the sole purpose of serialization. Do not use this id in other contexts.


leventeBajczi:
@csanadtelbisz, can you take a look at the conflicts?

csanadtelbisz:
I think this is not a crucial change for sv-comp, and it brings many fundamental changes (mostly refactoring that affects the core of Theta). So I vote for postponing this PR after sv-comp to avoid a hurried merge.

AdamZsofi:
Does this need work or discussion? (imho: if former, mark it as draft, if latter maybe discuss on Wednesday) @csanadtelbisz

csanadtelbisz:
We can close this for now. If we want a serializable class hierarchy in the future, we can quickly reproduce these changes (or do it better) with an LLM.

PR #393, opened 2025-10-06
Updated ltl properties for software witnesses

AdamZsofi:
I noticed not too long ago that we wrote `ERROR_LOCATION` to `unreach_call` witnesses. This is invalid (one should use the ltl specifications described in the SV-COMP rules).
Actually some validators did not care, but some do.

So I refactored the things around `ErrorDetection` to fix this and make it a bit more usable. 

Nicer solutions to the refactored code are welcome, otherwise please review and merge. :)


leventeBajczi:
@AdamZsofi please take a look at the merge conflicts!

csanadtelbisz (dismissed):
Sounds good to me!
Indeed, we need to use the `inputProperty` for witness generation, `verifiedProperty` is only for the inner working of Theta.
Other changes also look good.

PR #394, opened 2025-10-06
Bounded-like checker for LTS

kris7t:
This patch adds a bounded-like model checker that can work with LTS. This allows use to use BMC for models that can't be converted to `MonolithicExpr` directly, or when such conversion introduces a large overhead. Another benefit is testing new features (e.g., XCFA labels) that are only implemented in the LTS and `TransFunc` level, without having to provide a corresponding `MonolithicExpr` implementation.

* Adds `BoundedLtsChecker`, a checker that expands and LTS and an `Analysis` like an `Abstractor`, but applies a BMC-like procedure to call an SMT solver after every expanded transition and finally deliver a safe/unsafe/unknown verdict.
* In theory, the bounded-like checker could be used with any `Analysis` and `Prec`. Such a configuration would use the specified level of abstraction to enumerate transitions without passing the full path condition to the SMT solver before checking the full path condition like a BMC. However, we currently only use the unit abstraction, meaning that fireability is entirely determined by full path condition checking.
* Adds `UnitXcfaAnalysis`, an`Analysis` that uses the unit abstraction for XCFA. This analysis is not suitable for CEGAR, as it cannot be refined, but it lets `BoundedLtsChecker` handle the full path condition.


leventeBajczi:
Thanks for the PR! Besides a version bump, could you please also do the following so that this could be used from the CLI?
1. Create a new entry here: https://github.com/ftsrg/theta/blob/c7e1c3e0d6b40fef338a52086a4d3a6dbcfdf6b0/subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/params/ParamValues.kt#L66-L68
2. Add an entry here, with possible configuration options (create your own / reuse cegar / use `null` if no further input option is possible): https://github.com/ftsrg/theta/blob/c7e1c3e0d6b40fef338a52086a4d3a6dbcfdf6b0/subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/params/XcfaConfig.kt#L202-L227
3. Create a checker if that backend was specified (should probably be similar to cegar checker): https://github.com/ftsrg/theta/blob/c7e1c3e0d6b40fef338a52086a4d3a6dbcfdf6b0/subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/checkers/ConfigToChecker.kt#L48
4. Optionally extend these few lines to check if a safety proof is bounded, and return unknown if needed: https://github.com/ftsrg/theta/blob/c7e1c3e0d6b40fef338a52086a4d3a6dbcfdf6b0/subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/ExecuteConfig.kt#L264

This way it could be integrated easily into other parts of Theta, e.g., portfolios. Also, it would be easy to run a sanity check with SV-COMP tasks to see if anything was missed.

AdamZsofi:
What is the status of this PR? Should we close / update and merge / mark it as draft? @kris7t

leventeBajczi:
@kris7t could you please check if you agree with my modifications? 
I've tested the two problematic files and they seem fine now (I've added them to the integration tests as well, but those don't usually run with non-portfolio configs). 
Usually we don't take the coverage seriously for PRs, because cross-subprojects they are not counted properly. 
So, if you agree, I'd be happy to merge as-is (there aren't really modifications that would concern me with the rest of Theta anyway).

kris7t:
@leventeBajczi LGTM

PR #412, opened 2025-10-28
Xcfa multithread to monolithic

csanadtelbisz:
Adds a transformation from multithreaded XCFA to monolithic expression. Intended to be used by EmergenTheta at SV-COMP'26 for concurrent tasks.
Merge #391 before this PR.

### DIFF
--- a/subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/DereferenceToArrayPass.kt
+++ b/subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/DereferenceToArrayPass.kt
@@ -151,7 +151,7 @@ class DereferenceToArrayPass : ProcedurePass {
                         arrayType.elemType,
                       ),
                       cast(deref.offset.getArrayReads(xcfa), arrayType.elemType.indexType),
-                      cast(stmt.expr, arrayType.elemType.elemType),
+                      cast(stmt.expr.getArrayReads(xcfa), arrayType.elemType.elemType),
                     ),
                   ),
                   arrayType,


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Project" />
```
leventeBajczi: Is this deliberate?

csanadtelbisz:
The only main todo is running benchmarks. The dereference->array transformation was just completed by adding support for transforming multi-dimensional array literals in the (legacy) Z3 term transformer, so we need proper testing (the transformation of a Z3 model to Theta model failed so far).

As for the bounded portfolio update: I would not touch the bounded25 portfolio (it was for last year's svcomp), instead, we need a bounded26 portfolio. However, I would do it in a separate PR (where all portfolios are finalized)

leventeBajczi:
Some tasks (e.g., `sv-benchmarks/c/array-patterns/array10_pattern.c`) produce a `java.lang.IllegalStateException: Incomplete dereferences (missing uniquenessIdx) are not handled properly.` exception. This should not happen as dereferences are changed to arrays in my understanding -- could you please take a look @csanadtelbisz ?

leventeBajczi:
Test results (@mondokm, @csanadtelbisz): 

[datarace.table.html](https://github.com/user-attachments/files/23495842/datarace.table.html)
[V-COMP26_valid-memsafety.table.html](https://github.com/user-attachments/files/23495843/V-COMP26_valid-memsafety.table.html)
[SV-COMP26_unreach-call.table.html](https://github.com/user-attachments/files/23495845/SV-COMP26_unreach-call.table.html)
[SV-COMP26_termination.table.html](https://github.com/user-attachments/files/23495846/SV-COMP26_termination.table.html)
[SV-COMP26_no-overflow.table.html](https://github.com/user-attachments/files/23495847/SV-COMP26_no-overflow.table.html)

Shortlist of interesting stuff: 

* GSAT is really good with overflows
* dataraces are not handled by the monolithic conversion (couldn't we preprocess as with OC?)
* bitprecise arithmetic is not handled properly by whatever is handling `pthread_create` -- do we really not have anything in ConcurrencySafety where this matters? 
* IMC doesn't seem to be helped by using different solvers, even though with normal single-threaded non-heap programs, cvc5 far surpasses z3

leventeBajczi (approved):

PR #433, opened 2025-11-13
Libvsync

leventeBajczi:
- [x] Can parse non-variable references if they are referencing a literal-offset (currently only `0`) dereference
- [x] Can parse GNU C builtins 
- [x] Can parse ambiguous expressions such as `(len) & a` ( -- is it a cast? is it a bitwise AND? *i hate C*)
- [x] Can interpret GNU C builtins for atomic accesses


leventeBajczi:
@csanadtelbisz, if we were able to change the StartLabel/JoinLabel handles to expressions (i'm okay with constraining them to LHS expressions, i.e., RefExpr and DerefExpr), we could now parse libvsync tasks.

csanadtelbisz:
> @csanadtelbisz, if we were able to change the StartLabel/JoinLabel handles to expressions (i'm okay with constraining them to LHS expressions, i.e., RefExpr and DerefExpr), we could now parse libvsync tasks.

I do not see any obstacle: we would only need to add an extra case for creating a memoryassignment when necessary. I am expecting some changes in both the LTS-based analysis (XcfaState), OC checker event graph mapping and the monolithic conversion, but none of them should be too challenging. Shall I do it or are you working on it?

leventeBajczi:
Please do it if you have the time, I won't really be able to work on this.
I'll happily do the atomic builtin mappings if you'd like me to, though!

leventeBajczi:
Superseded by #524

PR #434, opened 2025-11-14
Wsl jar plugin

csanadtelbisz:
Adding a gradle plugin that can create a WSL-compatible JAR file. See more details among file changes in [doc/Build.md](https://github.com/ftsrg/theta/blob/4b5cac85d1250e349b87fdfaeb6632d498942ae8/doc/Build.md#building-and-running-theta-on-windows)

### DIFF
--- a/buildSrc/src/main/kotlin/WslJarPlugin.kt
+++ b/buildSrc/src/main/kotlin/WslJarPlugin.kt
@@ -41,34 +41,59 @@ class WslJarPlugin : Plugin<Project> {
     val wslJar = project.tasks.register("wslJar") {
       dependsOn(project.tasks.named("jar"))
 
-      if (System.getProperty("os.name").lowercase().contains("windows")) {
-        doLast {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/WslJarPlugin.kt
```diff
@@ -0,0 +1,113 @@
...
+  override fun apply(project: Project) {
+    if (!System.getProperty("os.name").lowercase().contains("windows")) return
+
+    if (!project.plugins.hasPlugin("application")) {
+      throw IllegalStateException(
+        "WslJarPlugin requires the 'application' plugin to be applied."
+      )
+    }
+
+    val wslJar = project.tasks.register("wslJar") {
```
arminzavada: It would probably be better to create a custom task type and use that here, instead of using a `doLast` lambda

leventeBajczi:
Thanks! I'll merge this into the source of #462, where we can continue the discussion.

PR #463, opened 2025-11-29
Fixed incorrect default trace-dir path

arminzavada:
The previous `inputOptions.model.parent` returns null if the model is directly in the working directory. Instead, the `parentFile` should be used, as that will respect relative paths as well.


AdamZsofi (dismissed):
Seems good. 
Honestly I have not tried, but if it works for you, it will work for me. I'll merge once the checks pass, sorry about the delay.

PR #464, opened 2025-11-30
Set entrypoint jar to absolute directory to allow customizing working dir

arminzavada:
Previous specification does not allow customizing the working directory in the container without breaking the default entrypoint.


AdamZsofi:
@leventeBajczi I'll approve and merge if you don't object. :)

PR #473, opened 2025-12-09
Fix simplification getting rid of variable type

### LINKED ISSUE #472: Termination: Unsound result with backend IMC?
Hi, thanks for fixing my previous bug,
I may have another one, this time with `--backend IMC`:

```
void f(int x){  // x == -1
  while (x < 0){
  }
}

int main() {
  unsigned int allOne = 0xffffffff;
  f(allOne);
  long castToLong = allOne;
  return 0;
}
```
Theta says this is a terminating-program, but in fact when the unsigned integer is casted to signed, this is a negative number and the loop does not terminate. 
It seems that there is an interplay with the unused variable `castToLong` as dropping this variable makes Theta return `SafetyResult Unsafe` as expected.

leventeBajczi:
Fixes #472


csanadtelbisz (dismissed):
LGTM, but please fix the tests before merging.

PR #483, opened 2026-03-03
Refactored to singleton logger

KlevisImeri:
# PR: Logger singleton 
Read [`Logging.md`](https://github.com/ftsrg/theta/pull/483/changes#diff-3ae987989a030d719c6e93b4670ab46db567ea7b500525e46e84a7e5bd7050a3)


subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/loopchecker/abstraction/GdfsSearchStrategies.kt
```diff
@@ -55,44 +53,39 @@ abstract class AbstractSearchStrategy : ILoopCheckerSearchStrategy {
...
-      Logger.Level.VERBOSE,
+    Logger.verbose(
       "Expanding through %s edge to %s node with state %s%n",
       if (incomingEdge.accepting) "accepting" else "not accepting",
       if (expandingNode.accepting) "accepting" else "not accepting",
       expandingNode.state,
     )
     if (expandingNode.state.isBottom()) {
-      logger.write(Logger.Level.VERBOSE, "Node is a dead end since its bottom%n")
+      Logger.verbose("Node is a dead end since its bottom%n")
```
leventeBajczi: if we're already changing this line, please fix the original typo (its -> it's / it is), even though I find the current literal meaning quite funny

subprojects/common/multi-tests/src/test/kotlin/multi/MultiNondetDiningPhilosophersTest.kt
```diff
@@ -140,7 +143,7 @@ class MultiNondetDiningPhilosophersTest {
         MultiPrec(cfaInitPrec, cfaInitPrec, dataInitPrec),
         dataInitPrec,
         Z3LegacySolverFactory.getInstance(),
-        logger,
+        Logger,
```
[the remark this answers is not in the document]
leventeBajczi: Mainly I don't understand why some places had their logger parameters removed, and then here some just receive the singleton object. I think we should opt for one of these solutions.

leventeBajczi (commented):
Thanks for the changes, I think these really help clear up the logging infrastructure of Theta!
I've skimmed through the changes, and these were my initial impressions. Please take a look -- but overall, I'm really happy with this PR.

subprojects/cfa/cfa-analysis/src/test/java/hu/bme/mit/theta/cfa/analysis/CfaTest.java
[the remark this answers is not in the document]
KlevisImeri: I wrote this above in the pull request text:
```Some other tests have been changes trying to fix issues which have to be reverted back like CFATest.```

As mentioned this is "half pull" request. I just wanted to get an approval if turning it into a singleton is something desirable, so know I can spend more time fixing every detail and making ti nice.

File: subprojects/common/common/src/main/java/hu/bme/mit/theta/common/logging/Logger.kt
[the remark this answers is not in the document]
Comment (KlevisImeri) [reply subprojects/common/common/src/main/java/hu/bme/mit/theta/common/logging/Logger.kt]: That is just me trying to use the logger and giving an example how I accept it to be used in the future. Yices is not special at all its just where I spend more time trying to find a workaround to the SSL expired certificate problem, and therefore I used the logger their more extensively so I can use it as a concreate example. The main idea is that eventually every class will have a type. (and also subtypes with in the that class type, is a tree of types with "grep" used to query it).

Lets suppose that I want to get the logs inside the `CegarChecker`. I would put as type:
- `INFO_CEGARCHECKER` -> create a func infoCegarChecker
- `MAINSTEP_CEGARCHECKER` -> create a func mainstepCegarChecker
- `SUBSTEP_CEGARCHECKER`  -> create a func substepCegarChecker
when I initialize with Logger.init("MAINSTEP|SUBSTEP") I would have both the `MAINSTEP_CEGARCHECKER`, `SUBSTEP_CEGARCHECKER`. But if I want only `CEGARCHECKER` then Logger.init("CEGARCHECKER") and it will show `SUBSTEP_CEGARCHECKER`, `MAINSTEP_CEGARCHECKER` and `INFO_CEGARCHECKER`.

Sorry I didn't do it justice explaining how it works. I will have to write better docs.

**Comment by mondokm** [comment]
Thank you for the PR, this is something that was needed for a while now:D

Levi commented above that he would like to see "new" gone from the CLI parameter descriptions. I think this is something that we should go even further with and completely avoid maintaining both a "new" and "old/legacy" logging system, and instead remove the legacy logging completely, including:
- the deprecated "Level" enum
- "LegacyLevel"
- "initOld"

If we reach a stable new logging system that can provide everything the legacy logging did, I think there is no need to maintain the legacy parts, they would only lead to confusion.

**Comment by KlevisImeri** [comment]
@AdamZsofi To implement the Legacy (old flags) I just translated them to the appropriate regex expressions for the "grep" API.

**Comment by AdamZsofi** [comment]
@KlevisImeri 
Can you please
- fix the conflicts
- check and "resolve conversation" in the above reviews if they are not relevant anymore? 

I'll do a second review on this in the meantime sometime.

File: doc/Logging.md
Comment (AdamZsofi) [inline doc/Logging.md:1]: Just to make it clear: log levels are still built upon each other, i.e., init to BENCHMARK will still log error, warn, etc., right?

File: subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/loopchecker/abstraction/GdfsSearchStrategies.kt (hunk 1 above)
Comment (AdamZsofi) [reply subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/loopchecker/abstraction/GdfsSearchStrategies.kt]: seems not to be fixed yet(?)

File: subprojects/common/common/src/test/kotlin/hu/bme/mit/theta/common/logging/LoggerTest.kt [hunk 3]
```diff
@@ -0,0 +1,277 @@
+/*
```
AdamZsofi, line 1: This is what I was looking for to understand the logger, great! Can you maybe reference these tests as examples in the doc?

subprojects/solver/solver-smtlib-cli/src/main/java/hu/bme/mit/theta/solver/smtlib/cli/SmtLibCli.java
```diff
@@ -34,15 +33,13 @@ public class SmtLibCli {
...
-    private Logger logger;
-
     static class MainParams {
 
         @Parameter(names = "--home", description = "The path of the solver registry")
         String home = SmtLibSolverManager.HOME.toAbsolutePath().toString();
 
-        @Parameter(names = "--loglevel", description = "Detailedness of logging")
-        Logger.Level logLevel = Logger.Level.MAINSTEP;
+        @Parameter(names = "--loglevel", description = "Regex pattern for log levels (e.g. ERROR|WARN|INFO)")
```
AdamZsofi: Okay, so, connecting to my first comment, I think I understand the logger itself now.
AFAIK if you say "INFO", it will ONLY display the info level logs, nothing else above/below - that is of course necessary to be able to set the granularity, regex, etc.

But maybe we misunderstood each other: on the clis, if I say "INFO" in `--loglevel`, I think I still actually want "INFO|everything below INFO", as it was before the refactor. I don't want to list all the levels. This is still the most common use case, so I don't think it can change.

(Am I correct with above or did I manage to misunderstand the whole thing?)

AdamZsofi: (This goes for all the other CLIs as well. I think some earlier reviews addressed this as well.)

AdamZsofi (changes requested):

doc/Logging.md
KlevisImeri: I removed that, but I can add it back. I just left it as a regex string. More simple.

subprojects/common/common/src/test/kotlin/hu/bme/mit/theta/common/logging/LoggerTest.kt
KlevisImeri: That is a great idea.

subprojects/solver/solver-smtlib-cli/src/main/java/hu/bme/mit/theta/solver/smtlib/cli/SmtLibCli.java
KlevisImeri: I can design such that if it says INFO i get only the "INFO|+everything below" but that means you have two user interfaces.
I mean how do you design the cli interface such that the user can ask both "INFO|+everything below" and just "INFO"?

The way maybe that it woudl be nicer is instead of calling `--loglevel` I woudl just call it `--log`. So for:
- "INFO|+everything below":  `./theta --log -level INFO`
- "INFO": `./theta --log "INFO"`

subprojects/common/common/src/main/java/hu/bme/mit/theta/common/logging/Logger.kt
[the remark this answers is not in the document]
KlevisImeri: I prefer having one file witch contains all the `logTypes` (option 2). I like it more for maintainability. Why? 
1. You can simply answer the question without searching the codebase: What logger types can I use? 
2. How can i use a logger type in many places in the codebase, in other words how do I pass the logger type around?

```
the central logTypes collection of all the log types is avoided and these values should be distributed, e.g., per class/per file. Something like STMT_LABELs and things like that - an extra "utility" constant. Pro: if I develop a new checker, I just add a constant or two and I instantly have a working, customized logger; (BIG) Con: extra constant value in practically every file
```
Not flexible enough. With other central types you can make whatever type you like it does not depend on the class or file, IT DEPENDS ON THE CONTEXT.

```
I think the other solution is to just keep logTypes, expect it to be huge and try to make that usable and maintainable - put it in it's own file and somehow make it foolproof in a way so that I don't forget to add my new log values to the new log checker
```
I think adding a new logger type is very simple. One constant + 1 boilerplate func.

```
Also, the question is: how much do we plan to fill up logTypes right now? Do we want to 1) go through a lot of places and add a lot of custom log types OR 2) do we do it gradually - e.g., you add what you needed (yices, cegar, whatever) and the next person who needs it elsewhere, say IC3, can add it there. I would prefer the latter - just with some really well shown / highlighted doc, so that people know that they can do this?
```
Doing it gradually is better cause the log types depend on the contex not the class so its not very simple to add, you have to understand that part of code. Therefore, refactoring all of it right now is like an "impossible" task.
Adding better docs is better.

PR #487, opened 2026-04-09
Made variable declarations optional

arminzavada:
There are cases, when the the variables are fully optimized away from the XSTS model, in which case the current implementation throws a parsing exception. This should not be the case however.


AdamZsofi:
@mondokm this seems like a minor change, can you take a quick look? We can update and merge afterwards.

mondokm:
> @mondokm this seems like a minor change, can you take a quick look? We can update and merge afterwards.

Yes, I bumped the version,  should be mergeable after all CI checks pass

mondokm (dismissed):

PR #495, opened 2026-05-05
Hardware

dantpu:
Refactored IC3Checker and added CarChecker and CarCegarChecker.

CarCegarChecker still depends on an earlier version of CarChecker. This will be fixed in the future.


subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/frame/base/ProofObligation.java
```diff
@@ -13,26 +13,22 @@
...
  *  limitations under the License.
  */
-package hu.bme.mit.theta.analysis.algorithm.ic3;
-
-import hu.bme.mit.theta.core.type.Expr;
-import hu.bme.mit.theta.core.type.booltype.BoolType;
-import java.util.Set;
+package hu.bme.mit.theta.analysis.algorithm.frame.base;
 
 public class ProofObligation {
```
AdamZsofi, line 18: Why proof _obligation_ , how is it more than a proof? (If it's "just" a proof, it could be connected to the Proof interface, etc.)

subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/frame/car/Node.java
```diff
@@ -0,0 +1,101 @@
...
+import hu.bme.mit.theta.core.type.booltype.BoolType;
+import hu.bme.mit.theta.core.type.booltype.FalseExpr;
+import hu.bme.mit.theta.core.type.booltype.SmartBoolExprs;
+import hu.bme.mit.theta.core.utils.PathUtils;
+import hu.bme.mit.theta.solver.UCSolver;
+import hu.bme.mit.theta.solver.utils.WithPushPop;
+import java.util.HashSet;
+import java.util.Set;
+
+public class Node {
```
AdamZsofi, line 33: Maybe call this something more specific, like FrameNode?

subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/frame/car/OverFrame.java
```diff
@@ -27,20 +30,21 @@
 import hu.bme.mit.theta.solver.SolverStatus;
 import hu.bme.mit.theta.solver.UCSolver;
 import hu.bme.mit.theta.solver.utils.WithPushPop;
-import java.util.*;
+import java.util.Collection;
+import java.util.Set;
 
-public class Frame {
-    private final Frame parent;
+public class OverFrame {
```
AdamZsofi, line 36: Over, as in over approximation or is this a more CAR specific concept?

AdamZsofi (changes requested):
Overall the code looks nice - I can not comment on algorithmic correctness.

My main concern is lack of docs, because this is the time to document *your* understanding of these algorithms and the architecture you built around it.

Ideas for docs:
- Document in local markdowns, e.g. a markdown in the frame directory, but a few class level javadoc comments wouldn't hurt either,
- link to the papers about the original algorithms,
- link to your thesis/tdk about the specific works.
- Explain what is what in the doc on a high level,
    - as you will see, I made some comments about not understanding some names, that's kind of the "architecture level" - what the building blocks are and how are, they connected,
    - explain the optimizations and add any conventions / rule of thumbs you know about - what you would (not) use in different situations; if there are any constraints (don't use this with that, etc.), write down those as well.

If I saw correctly, the algorithms so far are only connected to the (x)sts clis. I would really like to have it for xcfas, but that can be a separate PR. If you need help with it, just find me, please. 

_(I'll check why the CI is failing in more detail in a sec, but for formatting, run `./gradlew spotlessApply`.)_

subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/frame/base/ProofObligation.java
mondokm: This is called a proof obligation in IC3 literature, so I think we should stay consistent with that.
It's not connected to proofs in any way, it means a "state from the overapproximation, about which we have to prove if they are spurious or not"

AdamZsofi:
CI: I can see that you made version bumps, etc. 2 weeks ago - maybe we can just recheck and work out after changes were made and CI was re-executed, if the CI checks should be successful or not.

AdamZsofi:
@dantpu any open questions (i.e., are you blocked waiting for us) or is this just waiting for you to have time? :)

subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/frame/base/Cube.java
```diff
@@ -0,0 +1,78 @@
...
+import static hu.bme.mit.theta.core.type.booltype.SmartBoolExprs.And;
+
+import hu.bme.mit.theta.core.type.Expr;
+import hu.bme.mit.theta.core.type.booltype.BoolType;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public final class Cube {
```
mondokm, line 29: Please add a short javadoc comment that explains what a cube is

subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/frame/base/Clause.java
```diff
@@ -0,0 +1,75 @@
...
+import static hu.bme.mit.theta.core.type.booltype.SmartBoolExprs.Or;
+
+import hu.bme.mit.theta.core.type.Expr;
+import hu.bme.mit.theta.core.type.booltype.BoolType;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public final class Clause {
```
mondokm, line 29: Please add a short javadoc comment that explains what a clause is

subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/algorithm/frame/car/CarCheckerOld.java
mondokm, line 1: Does the current CarChecker work properly, or is this still needed?

mondokm (changes requested):
I think the PR is almost ready to be merged, only smaller questions remain (including the questions by @AdamZsofi ) as well as fixing the failing CI checks.

Would CAR and CARCEGAR also work on XCFA and STS in its current form? If yes, then I think it should also be introduced to XCFA and XSTS CLI.

Issue #506, opened 2026-06-09
Inlining doesn't instantiate local variables


s0mark:
The [procedure inlining pass](https://github.com/ftsrg/theta/blob/master/subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/InlineProceduresPass.kt) doesn't instantiate the inlined procedure's local variables. This can lead to unsound results, as evidenced by the following example:

https://github.com/ftsrg/theta/blob/ea0ee0f4ea84182706c3ce5e76abbdb00c17392c/subprojects/xcfa/xcfa-cli/src/test/resources/c/litmustest/singlethread/20testinline.c#L3-L12

The error location is reachable through the second call of `f(1)`, but Theta incorrectly gives a [safe verdict](https://theta.mit.bme.hu/?mode=C&config=eyJjb2RlIjoidm9pZCByZWFjaF9lcnJvcigpO1xuXG52b2lkIGZ1bmMoaW50IGEpIHtcbiAgICBpbnQgeDtcbiAgICBpZihhID09IDEgJiYgeCA9PSAyKSByZWFjaF9lcnJvcigpO1xuICAgIHggPSAxO1xufVxuXG5pbnQgbWFpbigpIHtcbiAgICBmdW5jKDApO1xuICAgIGZ1bmMoMSk7XG59Iiwic2VsZWN0ZWRQcm9wZXJ0eSI6InVucmVhY2gtY2FsbC5wcnAiLCJzZWxlY3RlZEV4YW1wbGUiOiIiLCJzZWxlY3RlZFJlbGVhc2UiOiJ2Ny4wLjMiLCJzZWxlY3RlZFByZXNldCI6IiIsImFyZ3MiOiItLWlucHV0ICVpbiAtLXByb3BlcnR5ICVwcm9wIC0tYmFja2VuZCBDRUdBUiAtLWVuYWJsZS14Y2ZhLXNlcmlhbGl6YXRpb24gLS1kb21haW4gUFJFRF9DQVJUIn0%3D), since it doesn't instantiate `x` and therefore the value from the previous call is reused.

leventeBajczi:
Thanks! 
Although I think this has come up before (maybe with the stack/recursion implementations?) and the verdict was, that because any such construct is UB, we don't care about this. 

[...]

PR #507, opened 2026-06-09
Dynamic local struct allocation

### LINKED ISSUE #474: Recursion and structs cause unsound results
[This](https://theta.mit.bme.hu/?mode=C&config=eyJjb2RlIjoiZXh0ZXJuIGludCByZWFjaF9lcnJvcigpO1xuXG50eXBlZGVmIHN0cnVjdCBBe1xuICBpbnQgaTtcbn0gX3N0cnVjdF9hO1xuXG52b2lkIGYoX0Jvb2wgaSkge1xuICBfc3RydWN0X2Egc3RyOyAvLyBub25kZXRcbiAgaWYoIWkgJiYgc3RyLmkpIHsgLy8gZmlyc3QgaW52b2NhdGlvbjogc3RyLmkgaXMgbm90IDBcbiAgICBmKDEpO1xuICB9XG4gIGlmKGkgJiYgIXN0ci5pKSB7IC8vIHNlY29uZCBpbnZvY2F0aW9uOiBzdHIuaSBpcyAwXG4gICAgcmVhY2hfZXJyb3IoKTsgLy8gc2hvdWxkIGJlIHJlYWNoYWJsZSwgdHdvIGluc3RhbmNlcyBvZiBfc3RydWN0X2EgZXhpc3RcbiAgfVxufVxuXG5pbnQgbWFpbigpIHtcbiAgZigwKTtcbn0iLCJzZWxlY3RlZFByb3BlcnR5IjoidW5yZWFjaC1jYWxsLnBycCIsInNlbGVjdGVkRXhhbXBsZSI6IiIsInNlbGVjdGVkUmVsZWFzZSI6InY2LjI3LjE0Iiwic2VsZWN0ZWRQcmVzZXQiOiIiLCJhcmdzIjoiLS1pbnB1dCAlaW4gLS1wcm9wZXJ0eSAlcHJvcCAtLWJhY2tlbmQgQ0VHQVIgLS1lbmFibGUteGNmYS1zZXJpYWxpemF0aW9uIC0tZG9tYWluIFBSRURfQ0FSVCJ9) example showcases an example where we don't track the stack-allocated struct base address well across recursive calls.

s0mark:
Makes local variable struct address allocation happen dynamically during verification, as opposed to statically in the frontend phase. Fixes #474.

The key challenge to overcome here is to figure out whether or not a local variable is a compound type when a procedure is called. From my understanding, this information is only tracked in the frontend and cannot be retrieved from a `VarDecl` or its `.ref` expression without some `parseContext`.

This solution adds optional initializers to all local variables of an `XcfaProcedure`, which can be called to generate an initializing expression. When a procedure call is encountered, the generated expression is assigned to the new instance of the local variable, at the same time when parameters are assigned. The initializers are added in the frontend phase and are only used for this purpose at this point.

Another approach I considered was to collect such local variables in a procedure pass and store them in a some heap class, along with their `CComplexType` (for the address type). Upon a procedure call, this collection could be queried to retrieve the initializing memory address for a local variable, if required. I decided on the former solution, as it seemed more cohesive.

If there's a preferred, perhaps less intrusive way of retrieving the necessary information, please let me know!


leventeBajczi (commented):
I'm afraid this breaks non-CEGAR approaches -- even though they would fail anyway due to `dereference` not being translated, that can/will be fixed at some point, and we'll get silent failures/wrong results for non-initialized struct variables.

The original problem is, as far as I understand it, that we used a fix pointer base value for the local struct. Then, in a recursive call, we did not properly 'reallocate' it, but kept using the same base value.

The trivial solution to me would be to get rid of stack-allocation here (it _is_ unsound, after all) and divert it to the same machinery as what handles `malloc()` and co. What do you think?

s0mark:
> I'm afraid this breaks non-CEGAR approaches -- even though they would fail anyway due to dereference not being translated, that can/will be fixed at some point, and we'll get silent failures/wrong results for non-initialized struct variables.

That's a good point. How do other approaches handle (non-inlined) procedure calls though, especially parameter assignment? If they use the [code](https://github.com/s0mark/theta/blob/b066a208f8c836fbbec1e020bf60c9c1e514f42b/subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/XcfaState.kt#L342-L402) from `XcfaState`, they should also handle this.

> The original problem is, as far as I understand it, that we used a fix pointer base value for the local struct. Then, in a recursive call, we did not properly 'reallocate' it, but kept using the same base value.

Yes, although it's not a direct result of recursion, but a non-inlined procedure. The issue is that the address calculation happens statically in the frontend phase.

> The trivial solution to me would be to get rid of stack-allocation here (it is unsound, after all) and divert it to the same machinery as what handles malloc() and co. What do you think?

I like the simplicity of the idea. I believe it could solve both #474 and #506. I do have some concerns:

- Wouldn't this affect our memory safety analysis in an unwanted manner? I'm not sure exactly how it works currently, but I could see some unintended side effects, e.g. a `free` might need to be added at the end of the function in order not to violate the memtrack property.
- It wouldn't be sound straight away, as the malloc handling isn't atomic, therefore, this example from #506 with malloc doesn't get verified correctly. That could be fixed by wrapping it in an atomic block though.
- Currently, the local struct generation uses $$3k + 1$$ addresses, while malloc uses $$3k$$. I'm not sure if we rely on this anywhere, but in case we do, perhaps it'd make sense to introduce a separate `__malloc` variable for local initializations.

s0mark:
#524 uses `__malloc` for local struct addresses and fixes #474. Closing this PR in favour of that one.

PR #511, opened 2026-06-17
Points-to graph unbox fix with pos

csanadtelbisz:
Implementing a fix for points-to graph calculation.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -47,7 +47,7 @@ buildscript {
 
 allprojects {
     group = "hu.bme.mit.theta"
-    version = "7.2.0"
+    version = "7.2.1"
 
     apply(from = rootDir.resolve("gradle/shared-with-buildSrc/mirrors.gradle.kts"))
 }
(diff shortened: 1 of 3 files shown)


leventeBajczi (dismissed):
Thanks for the fix!
If you wouldn't mind, could you add the problematic testcase to the set of integration tests? Just adding the yaml, the source, a property, and a `<tasks>` tag into [this file](https://github.com/ftsrg/theta/blob/master/integration-tests/software/theta.xml).

csanadtelbisz:
Adding this file alone seems a bit ad hoc to me. We should probably add a large proportion of the goblint regression tests which is a test suite with the exact same purpose containing mostly very simple tasks.

leventeBajczi:
Sure, if you have a good idea for a subset that would test data races thoroughly enough, I'm happy for it -- my idea was just to not got looking for tasks now, since we have a good candidate for a 'tricky' (at least for us) task from the issue. To be honest, I just realized we didn't have any tests for dataraces in the test suite, hence the request to add at least one.

leventeBajczi (approved):

csanadtelbisz:
IMO, we could do that in a separate PR.

leventeBajczi:
Sure! Will you, or should i?

PR #519, opened 2026-07-03
Archive timestamp fix

s0mark:
Fixes the issue with files in the artifact having their timestamps removed. Adds the options to configure what platform and architecture the artifact should be built for, using the  `-Dos.name` and `-Dos.arch` flags with the corresponding `buildArchive` task.


leventeBajczi (approved):
Great work, thanks!

leventeBajczi:
Please bump the version, then feel free to merge :)

PR #523, opened 2026-07-10
Saturation + CEGAR

mondokm:
Saturation-based CEGAR (Different from the checker-agnostic MonolithicExpr cegar that wraps checkers via the predicate abstraction pass).

- MDD expression nodes and next-state descriptors reworked: per-graph caches and strategies, lazy traversers return their solver after single-shot queries, explicit-representation extraction, postcondition (target-only) descriptors and their intersection.
- Trace generation shared by `MddChecker` and `MddCegarChecker`, with `--trace-search DFS|BFS|BFS_BACKWARD` (the breadth-first variants give a shortest counterexample); every step carries the transition that fired.
- `--memory-init ZERO|UNCONSTRAINED|AUTO` for the dereference memory array (AUTO is zero for the MDD backend, whose fixpoint needs a finite set of initial states).
- SMT-LIB: an unparsable `(get-model)` reply is reported instead of a bare `AssertionError`; `(as const ...)` with a non-literal element becomes an `ArrayInit` instead of failing.
- Updated `hu.bme.mit.delta` library jar.


subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/params/ParamValues.kt
```diff
@@ -96,13 +87,13 @@ enum class Backend {
...
   IMC,
   KINDIMC,
-  PATH_ENUMERATION,
   CHC,
   OC,
   LAZY,
   PORTFOLIO,
   TRACEGEN,
   MDD,
+  MDD_CEGAR,
```
leventeBajczi, line 96: Is this a separate enough backend that it warrants a new entry here? Usuaually, implicit predicate abstraction is a flag under the corresponding backend.

subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/params/XcfaConfig.kt
```diff
@@ -623,6 +602,59 @@ data class MddConfig(
   var initPrec: InitPrec = InitPrec.EMPTY,
 ) : SpecBackendConfig
 
+data class MddCegarConfig(
```
leventeBajczi, line 605: if this becomes a sub-setting of MDD rather than its own block, then MDD can still wrap it directly, we don't need to merge the two configs if we don't want to.

leventeBajczi (approved):
Thanks for the new backend, looks promising!
I'm happy to approve, I just left two minor comments for the xcfa part of the PR. 
Feel free to resolve it without modification if you think they're not worth changing.

leventeBajczi:
Also, don't forget that the PR description will become the release note as well, so maybe it's worth removing the "This PR" part.

mondokm:
> Also, don't forget that the PR description will become the release note as well, so maybe it's worth removing the "This PR" part.

I intended to publish this PR as a draft for now, but I will update the description for the merge-ready version

subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/params/ParamValues.kt
mondokm: Yes, it is very different from the simple algorithm-agnostic implicit wrapper. For now, I need both as separate CLI bindings to be able to benchmark them against each other, but later I will consider placing this under the flag you mentioned (although it would be somewhat confusing that way, because elsewhere the flag implies the simple wrapper).

PR #524, opened 2026-08-25
Svcomp27 fixes (🤖)

### LINKED ISSUE #474: Recursion and structs cause unsound results

leventeBajczi:
# SV-COMP 2027 frontend and soundness work

Gets theta's C frontend through the SV-COMP benchmark suite, and makes the answers it
produces trustworthy.

## Measured impact

Full portfolio, 36,531 tasks, 15 min / 15 GB / 2 cores, against the released v7.3.1:

|                  |   v7.3.1 | this branch |
|------------------|---------:|------------:|
| score            |    8,071 |  **22,536** |
| correct          |    6,017 |  **14,866** |
| wrong            |      147 |      **72** |
| tool errors      |   16,194 |   **2,337** |
| timeout / OOM    |   14,136 |      18,839 |

Tool errors are down 86%. Resource exhaustion rose *because* far more tasks now run far
enough to consume resources instead of being refused in seconds.

Two families the release could not read at all now parse completely: intel-tdx-module
(836 tasks, frontend errors 836 to 0) and libvsync (52 to 0). Neither solves anything
yet; both are analysis-capacity limited. That was measured, not assumed.

## What changed

- **Frontend.** Struct tags cached while their own body was still being read; lvalue
  shapes (a bitfield reached through a narrowing, a struct copied through a pointer,
  aggregate assignment, brace elision); extern arrays of unspecified extent; union cell
  widths.
- **Library modelling.** calloc, symbolic memset, va_arg, variadic arity, pthread_detach,
  and stubs for the stdio/string functions nothing defines - each stub's havoc bounded to
  the C type it writes.
- **Encodings.** Subnormal float decoding; INT_MIN negation in the bitvector overflow
  check; a byte-addressed memory model, with a fallback chain multi to flat to bytes for
  programs the cell-per-value models cannot express.
- **Diagnostics.** Error messages now name the C type, the declaration or the fields.

## Validation

- **Canary suite**: 268 real SV-COMP tasks + 73 feature-guard fixtures, green on every
  commit that ships. Registered as a Gradle task; see subprojects/xcfa/xcfa-cli/canaries.
- **Unit tests**: 1,093 green.
- **A/B for every fix**, rebuilding both sides. Three changes were implemented, measured
  at zero effect, and reverted rather than shipped.

## Known limitations

1. **Wrong answers rose 55 to 72.** 68 of the 72 were errors in the release: unsoundness
   newly exposed by tasks that now get far enough to answer, not caused here. Two went
   correct to wrong - ldv-regression/rule60_list2 (a pre-existing bitvector defect, newly
   reached) and pthread-divine/tls_basic (not diagnosed).
2. **Best-localised soundness target**: the false alarm on libvsync
   bounded_mpmc_check_full, reproduced independently by three configurations, which puts
   the defect in race detection itself rather than in one domain or POR setting.
3. **One fix works but is unexplained.** The library-stub range bound turns a family of
   overflow false alarms correct on real inputs, yet minimal repros verify the same with
   and without it. Documented at the call site; do not remove it on the strength of a
   small repro.
4. **8 ntdrivers tasks still refuse** an element-to-element struct copy. The cause is
   known and the obvious fix regressed other tasks.
5. **The byte-addressed model refuses floats.** SMT-LIB's FP sort has a single NaN, so the
   IEEE round trip cannot preserve a payload. A loud refusal beats a wrong answer.

## Note for reviewers

This branch also fixes a command injection in the release workflow: two steps interpolated
the release message into a shell script, so backticks in any PR description were executed
on the runner. Both now pass their values through the environment.


subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/ptr/PtrPrec.kt
csanadtelbisz, line 1: Are these changes to the copyright year intentional? If it is not, please revert them as there is no other change in these files and setting a copyright year backwards is at least suspicious.

Same for other similar changes.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
```diff
@@ -93,9 +93,43 @@ internal class XcfaToEventGraph(private val xcfa: XCFA, private val parseContext
...
+   * so accesses of different types never communicate there either. Keeping the OC encoding in sync
+   * means the OC backend makes the same aliasing assumptions as the CEGAR backends rather than
+   * inventing new ones.
+   *
+   * Note that the partitioning cannot hide a data race: races are detected by the instrumentation
+   * of `DataRaceToReachabilityPass` (ordinary flag variables holding the racing address), not by
+   * the read-from relation. Splitting only removes read-from edges, and since every partition keeps
+   * its own unconstrained "garbage" initial write, a read that loses a cross-type source falls back
+   * to an arbitrary value -- i.e. the encoding gains behaviours, never loses them.
+   */
```
csanadtelbisz, line 102: Please shorten this comment (and comments in general in files that are potentially developed by humans in the future). The second and last paragraphs are not very informative, the first and third should be enough - and even those could be compressed.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaOcChecker.kt
```diff
@@ -111,7 +112,23 @@ class XcfaOcChecker(
    */
   private fun check(forceUnrollBound: Int): Pair<SafetyResult<EmptyProof, Cex>, Boolean> {
     // force loop unroll for BMC
-    val xcfa = xcfa.optimizeFurther(ProcedurePassManager(listOf(LoopUnrollPass(forceUnrollBound))))
+    // UnusedLocRemovalPass after the unroll: force unrolling leaves behind copies past the bound
```
csanadtelbisz: If this is general behavior that the loop unroll pass leaves some leftover elements in the CFA then we should rather fix that pass: best would be properly fixing it, that is, the loop unroll logic should not leave unused remnants in the CFA, but if we do not have the capacity for that, running this unused loc removal pass at the end of loop unroll pass is still better then needing to add it manually anywhere loop unroll is used.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
```diff
@@ -109,11 +143,71 @@ internal class XcfaToEventGraph(private val xcfa: XCFA, private val parseContext
...
 
+  /**
+   * Collects every dereference type triple occurring in the XCFA and creates a memory declaration
+   * for each.
+   *
+   * This has to happen up front (rather than lazily, on first access): every partition needs an
+   * unconstrained initial write that is program-order-before all other events, and such an event
+   * can only be created while the entry thread is being set up. A read with no read-from source at
+   * all would be left with a completely free value, which would be unsound.
+   */
```
csanadtelbisz, line 137: Again, compress comment, please.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
```diff
@@ -109,11 +143,71 @@ internal class XcfaToEventGraph(private val xcfa: XCFA, private val parseContext
...
+        Decls.Var(name, key.elemType)
+      }
+  }
+
+  /**
+   * `pthread_{get,set}specific` and `pthread_key_create` are modelled with dereferences synthesised
+   * in [ThreadProcessor.process] that are not present in the XCFA itself, so their partitions have
+   * to be registered separately.
+   */
+  private fun pthreadSpecificKey(label: InvokeLabel): MemoryKey? {
```
csanadtelbisz: Just flagging that we should probably add proper thread-specific variable support on the XCFA label. Then we would not need such specific hacks in the analyses. It is not trivial and that is why it was hard-coded here in the event graph creation in the first place, but it gets uglier with each such addition.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
```diff
@@ -109,11 +143,71 @@ internal class XcfaToEventGraph(private val xcfa: XCFA, private val parseContext
...
+        "pthread_key_create" ->
+          ((label.params.getOrNull(1) as? RefExpr<*>)?.decl as? VarDecl<*>)?.type
+
+        else -> null
+      } ?: return null
+    return MemoryKey(keyType, Int(), Int())
+  }
+
+  /** Alphanumeric rendering of the triple (const names must stay parseable, see reason parser) */
+  private val MemoryKey.suffix: String
```
csanadtelbisz: Please add this as a member method. I see no point in defining it as an extension method.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
```diff
@@ -109,11 +143,71 @@ internal class XcfaToEventGraph(private val xcfa: XCFA, private val parseContext
...
+    get() =
+      listOf(arrayType, offsetType, elemType).joinToString("_") {
+        it.toString().replace(Regex("[^A-Za-z0-9]"), "")
+      }
+
+  private fun memoryDeclOf(deref: Dereference<*, *, *>): VarDecl<Type> =
```
csanadtelbisz: On the other hand, following the (unwritten and not really crucial) conventions in the `oc` package should be an extension property as it has no further parameters: `Dereference<*, *, *>.memoryDecl`.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
```diff
@@ -337,9 +447,102 @@ internal class XcfaToEventGraph(private val xcfa: XCFA, private val parseContext
...
+        // copies past the unroll bound that nothing can reach. An edge out of such a location can
+        // never fire, so waiting for it would strand a perfectly ordinary merge point forever --
+        // that is the "dangling edges" half of the old error. Let those items through with the
+        // predecessors they did get. A real loop is unaffected: its head is blocked by its own
+        // back edge, whose source is reachable from the head itself, so it is never releasable and
+        // still reports below.
+        if (toVisit.isEmpty() && waitList.isNotEmpty()) {
+          val releasable = releasableFrom(waitList, visited)
+          if (releasable.isEmpty()) {
+            if (System.getenv("THETA_OC_LOOP_DEBUG") != null) {
```
csanadtelbisz: I am not sure we need these changes starting at this line. It seems like an AI debug leftover. Probably a larger scope of changes is not required at all (the whole releasedEarly stuff?). Please clean this part (basically anything that is related to these releasable/releasedEarly, etc. variables. Probably, this is solved by UnusedLocRemovalPass (see other changes and my other comment).

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/XcfaDataRaceCheck.kt
```diff
@@ -58,10 +59,13 @@ data class DataRace(
...
- * both atomic, not mutually excluded) enabled after the given state, or returns null if no data
- * race is possible.
+ * both atomic, not mutually excluded) enabled after [s], or null if no data race is possible.
+ *
+ * The returned [DataRace] carries the racing edges and labels the concurrent-witness writer needs;
+ * [getDataRaceDetector] and the trace-checker wrapper only look at [DataRace.condition]. `_Atomic`
+ * accesses are excluded here (via [parseContext]), so both consumers stay atomic-aware.
  */
-fun findDataRace(s: XcfaState<out PtrState<out ExprState>>): DataRace? {
+fun findDataRace(s: XcfaState<out PtrState<out ExprState>>, parseContext: ParseContext): DataRace? {
```
csanadtelbisz, line 70: I do not like the idea of needing a ParseContext here. I understand that we need it for deciding if a dereference points to atomic data. But then, we should probably encoded this in the Type (e.g., having an Atomic wrapper type or something like this). This is too much frontend magic in a backend analysis.

subprojects/xcfa/xcfa-cli/canaries/run_fixtures.sh
csanadtelbisz, line 1: Having this as a gradle script might be a better solution to keep everything as cross-platform as possible.

subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/params/XcfaConfig.kt
```diff
@@ -136,6 +136,12 @@ data class FrontendConfig<T : SpecFrontendConfig>(
       "Number of loop iteration to unroll even if the number of iterations is unknown; in case of such a bounded loop unrolling, the safety result cannot be safe (use -1 to disable)",
   )
   var forceUnroll: Int = -1,
+  @Parameter(
+    names = ["--no-force-unroll-recursion"],
+    description =
+      "Disable expanding recursive procedure calls to the depth loops are force-unrolled to. Expansion is on wherever a force-unroll bound is in effect: calls still recursive at that depth are cut, so as with force unrolling the safety result cannot be safe, and it lets backends that need a call-free CFA (e.g. OC) handle programs whose recursion depth is bounded.",
+  )
+  var noForceUnrollRecursion: Boolean = false,
```
csanadtelbisz: We should probably have a separate `recursionForceUnroll` option and use it whereever needed, instead of disabling the effect of another option in certain cases. If there is no better reason for having it this way.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/Builders.kt
```diff
@@ -152,6 +153,7 @@ constructor(
     var that = if (this::partlyOptimized.isInitialized) partlyOptimized else this
     for (pass in manager.passes[phase]) {
       that = pass.run(that)
+      that.checkEdgesHaveLocations(pass)
```
csanadtelbisz: This is probably not the best place to add this check as we can run passes anywhere (and we actually do so in Theta). So we should probably have a pass runner that runs the pass and executes the check while keeping the actual run method of the pass private. Or at least, it could be implemented as a method on procedure pass manager: `manager.run(pass, that)` or something like this as passes are often executed using a manager..

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/Builders.kt
```diff
@@ -277,6 +288,39 @@ constructor(
...
+  private fun checkEdgesHaveLocations(pass: ProcedurePass) {
+    // Identity, not equality. XcfaLocation is a data class, so a *different instance* carrying the
+    // same name/flags/metadata compares equal and satisfies `in locs` -- while owning its own,
+    // separate incoming/outgoing sets. An edge attached to such a stray twin is invisible to every
+    // traversal that walks adjacency (which is all of them, including LoopUnrollPass's back-edge
+    // cut), yet XcfaProcedure.deepCopy resolves endpoints through a map keyed by equality and so
+    // silently re-points the edge onto the registered instance. A cycle hidden that way only
+    // materialises in the copy, where it surfaces as the OC checker rejecting the task for "loops".
+    val registered =
+      java.util.Collections.newSetFromMap(java.util.IdentityHashMap<XcfaLocation, Boolean>())
```
csanadtelbisz: Since XcfaLocation is a data class, I do not really understand why identity hash map is necessary here. Can you give an example? Do we ever rely on actual instances? We should probably change that instead of using an identityhashmap here.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/XCFA.kt
```diff
@@ -188,7 +188,19 @@ constructor(
...
+   * The *object this points at* is `_Atomic`: `_Atomic int *p`, or the pointer that
+   * [ReferenceElimination] makes for an address-taken `_Atomic int`.
+   *
+   * A different question from [atomic], and the one a *memory* access has to ask. It is recorded
+   * here rather than looked up when needed because the C types are keyed by object identity, so a
+   * pass that rebuilds an expression loses them -- and by the time the analysis runs, an
+   * address-taken object's pointer has been folded to a bare literal (its id) with nothing left to
+   * read.
+   */
+  val pointsToAtomic: Boolean = false,
```
csanadtelbisz, line 203: If this is included in the model, then why do we need the complicated parsecontext-based stuff in data race check? (see other comment)

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/AllocaFunctionPass.kt
```diff
@@ -0,0 +1,244 @@
...
+ * Known gaps (both are the pre-existing scope-lifetime limitation, not new to alloca): the block is
+ * never invalidated at function return, so a dangling access to it afterwards is not caught, and
+ * `free()`ing it is accepted rather than reported as an invalid free.
+ *
+ * Requires the ProcedureBuilder be `deterministic`.
+ */
+class AllocaFunctionPass(val parseContext: ParseContext) : ProcedurePass {
+
+  override fun run(builder: XcfaProcedureBuilder): XcfaProcedureBuilder {
+    val mallocVar = builder.parent.mallocVar(parseContext)
```
csanadtelbisz, line 63: If we use the malloc var and this util in several files, move it to a util class instead of importing it from the companion object of the other class.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/AllocaFunctionPass.kt
```diff
@@ -0,0 +1,244 @@
...
+import hu.bme.mit.theta.xcfa.model.*
+import hu.bme.mit.theta.xcfa.passes.MallocFunctionPass.Companion.ensureMallocVar
+import hu.bme.mit.theta.xcfa.passes.MallocFunctionPass.Companion.firstAllocationRetType
+import hu.bme.mit.theta.xcfa.passes.MallocFunctionPass.Companion.mallocVar
+import hu.bme.mit.theta.xcfa.utils.AssignStmtLabel
+import hu.bme.mit.theta.xcfa.utils.getFlatLabels
+
+/**
+ * Turns `alloca(size)` into an address assignment, like [MallocFunctionPass] does for `malloc`, but
+ * places the block in a different residue class of the pointer base space.
```
csanadtelbisz, line 39: I cannot really follow the pointer base space classes. Can you collect that somewhere? We have one for malloc, for alloc now, for reference elimination; what else? Are we sure that they never overlap?

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/CallocFunctionPass.kt
```diff
@@ -0,0 +1,146 @@
...
+import hu.bme.mit.theta.core.utils.ExprUtils
+import hu.bme.mit.theta.frontend.ParseContext
+import hu.bme.mit.theta.frontend.transformation.model.types.complex.CComplexType
+import hu.bme.mit.theta.xcfa.model.*
+import hu.bme.mit.theta.xcfa.utils.getFlatLabels
+import java.math.BigInteger
+
+/**
+ * `calloc`, which nothing modelled: it reached the analysis as a call to a procedure that does not
+ * exist and brought it down with "No such method calloc".
```
csanadtelbisz: These are perfect examples of "who cares?" historical comments produced in large amount.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/DataRaceToReachabilityPass.kt
```diff
@@ -63,21 +73,67 @@ class DataRaceToReachabilityPass(private val property: XcfaProperty, enabled: Bo
...
+        is IntType -> Int(-1) as LitExpr<T>
+        is BvType ->
+          BvUtils.bigIntegerToSignedBvLitExpr(BigInteger.valueOf(-1), type.size) as LitExpr<T>
+        else -> error("Cannot detect races on dereferences with component type $type")
+      }
+    }
 
-    private val Expr<*>.derefOffsetWriteFlag: VarDecl<IntType>
-      get() = derefOffsetWriteFlagVar.also { check(type == Int()) }
+    private fun clearFlag(flag: VarDecl<*>): XcfaLabel = AssignStmtLabel(flag, noAccess(flag.type))
```
csanadtelbisz, line 108: I already have a fix for this pass that uses counters instead of flags that is necessary when using read-write flags (multiple threads can read-lock on the same variable). It seems that it was not on master. It is an easy fix, we could maybe change it here: setting the flag is an increment and clearing a flag should be a decrement, that's all.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/DereferenceToArrayPass.kt
```diff
@@ -58,63 +40,44 @@ private typealias ArrayType2D = ArrayType<out Type, ArrayType<out Type, out Type
...
+ * inconsistent values for them (false counterexamples on ldv-regression, among others). The array
+ * starts havoced: stack and heap cells are garbage until written, and globals do not need the
+ * array's default value because their zero/explicit initialization is materialized as ordinary
+ * writes in the init procedure.
  */
 class DereferenceToArrayPass : ProcedurePass {
 
-  private lateinit var arraysByType:
-    Map<Tuple4<Type, Type, Type, Boolean>, VarDecl<out ArrayType2D>>
+  private lateinit var arraysByType: Map<Triple<Type, Type, Type>, VarDecl<out ArrayType2D>>
```
csanadtelbisz: Since this type key is such a common theme (see the changes in oc checker and maybe some other places), we should move it to a util file with the accompanying methods.

I prefer having a deticated class with meaningful attribute names instead of Triple here.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/LibraryStubsPass.kt
```diff
@@ -0,0 +1,265 @@
...
+ * THROUGH, and those pointees are filled with fresh nondeterministic values rather than being left
+ * at their old contents -- leaving them would be a specific wrong value, which can hide a bug as
+ * easily as invent one, where unconstrained is a safe over-approximation.
+ *
+ * ⚠️ Not modelled here, deliberately:
+ * - a call flagged [InvokeLabel.isLibraryFunction]. That flag means "an later pass or the analysis
+ *   handles this one specifically" -- [CLibraryFunctionsPass] sets it on the thread-specific-key
+ *   family, which the OC checker supports properly. Stubbing such a call to a havoc throws that
+ *   support away, so this pass skips them exactly as [UnresolvedInvokeToHavocPass] does.
+ * - anything an earlier pass already consumes ([CLibraryFunctionsPass] takes `printf` and `scanf`,
```
csanadtelbisz, line 57: I am not sure if the implementation in `CLibraryFunctionPass` adds any value to printf and scanf beyond the stubs here. Please check, and move printf and/or scanf here, if there is no extra value in having the implementations of prtinf and/or scanf there.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/UnrollPass.kt
```diff
@@ -38,13 +43,52 @@ import kotlin.math.max
...
-class LoopUnrollPass(alwaysForceUnroll: Int = -1) : ProcedurePass {
+/**
+ * @param substituteLoopVar when true, each unrolled copy has the loop variable replaced by its
+ *   constant value for that iteration (`&t[i]` becomes `&t[0]`, `&t[1]`, …). Only the loop variable
+ *   is substituted, so address-of expressions of other variables (`&x`) are left for
+ *   ReferenceElimination -- which is why this may safely run before it. Requires [parseContext].
+ */
+class LoopUnrollPass(
+  alwaysForceUnroll: Int = -1,
+  private val substituteLoopVar: Boolean = false,
```
csanadtelbisz, line 64: 1. This should probably be the task of SimplifyExprsPass to do this whenever it can be automatically inferred. If it was not done already, then probably the expression simplifier implementation is not wide enough (e.g., not implemented well for dereferences if &t[i] is not simplified to &t[0]).
2. Is `substituteLoopVar = true` sound for data races/memory safety (when the loop var is not local)? If not, please ensure that we do not use this option in that case. Either as a config option validation rule, or as a check here.
3. If substituting a loop var is sound, why not always use it?

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/UnrollPass.kt
```diff
@@ -190,14 +306,142 @@ class LoopUnrollPass(alwaysForceUnroll: Int = -1) : ProcedurePass {
...
+   * call in that procedure un-inlined, not just the recursive one. Backends that need a call-free
+   * CFA (the OC checker does) then reject the task outright. Expanding here recovers those programs
+   * whenever the interesting depth is bounded -- and, because this runs per force-unroll bound
+   * rather than once at inlining time, raising the bound genuinely re-expands the recursion deeper.
+   *
+   * A call that is still recursive at the bound is cut, dropping the executions past it, which is
+   * the same promise force unrolling makes for loops; [XcfaProcedureBuilder.setUnsafeUnroll]
+   * records that so a `safe` verdict stays flagged as bound-limited.
+   */
+  private fun unrollRecursiveCalls(builder: XcfaProcedureBuilder) {
```
csanadtelbisz, line 346: We should probably rename the class to reflect the changes (not only loops are unrolled anymore): e.g., UnrollPass.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/ProcedurePassManager.kt
csanadtelbisz, line 1: Excessive comments are particularly annoying in this file. We should use pass class names that are descriptive enough to give an idea what they do.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/README.md
csanadtelbisz, line 1: Having such a file that summarizes passes is a good idea, but this one is rather "ad hoc", only mentioning a few aspects of a few passes. Currently, this is more of a changelist and I would not commit this to master as is.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/ReallocFunctionPass.kt
csanadtelbisz, line 1: Does this need to be a separate pass? Cannot it be integrated into another existing pass? The list of passes grows quite rapidly and will soon reach the level of comfortable human readability in my opinion. What about putting this in CLibraryFunctionsPass or AllocaFunctionPass?

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/TypeRange.kt
csanadtelbisz, line 1: Could not this go to an existing XCFA util class? It is somewhat surprising to see such a file in the `passes` package.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/WitnessOptimizer.kt
csanadtelbisz, line 1: Could not this file be merged with ApplyWitnessPass?

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/utils/AtomicAccessUtils.kt
csanadtelbisz, line 1: See my other comment regarding data race check: I do not like needing the parsecontext for deciding something that changes verification behavior/verdict. It should be part of the model instead.

csanadtelbisz (changes requested):
I left quite some comments here-and-there. The major notes/requests:
- The excessive comments are rather annoying in files that are intended to be developed by humans in the future (i.e., mostly everything regarding the XCFA and analyses). Please reduce/compress these comments, and maybe update CLAUDE.md to instruct the LLM to write short comments. Frontend classes can stay as they are as we do not really like to touch them anyway.
- ParseContext is injected too often. Everything that affects the analysis should be part of the XCFA (either as some specific entry in the XCFA model or Theta core elements such as experssions as types). Currently, this is mostly manifested by the referenced atomic types.

CLAUDE.md
mondokm, line 1: These benchmarking instructions are specific to @leventeBajczi 's environment, this file should not be committed to master

.gitignore
```diff
@@ -39,3 +39,14 @@ xcfa-*.smt2
...
+# Python bytecode from the canary helper scripts
+__pycache__/
+*.pyc
+
+# IDE settings are personal, not project configuration
+.vscode/
+
+# Benchmark runs, triage notes and result archives are kept locally, not in the repo.
+# The canary suite that gates changes lives in subprojects/xcfa/xcfa-cli/canaries.
+benchmark-results/
```
mondokm: The `benchmark-results` folder and this entry here is an artifact of the Claude workflow, should be removed from the PR. Also, the comments in this file are unnecessary

subprojects/xcfa/xcfa/src/test/java/hu/bme/mit/theta/xcfa/passes/UnresolvedInvokeToHavocTest.kt
mondokm, line 1: This test seems like a Claude debugging artifact. We do not have such tests for other passes, we probably don't need this.

subprojects/xcfa/xcfa-cli/README.md
```diff
@@ -182,3 +182,18 @@ XcfaConfig            "1" *-[#595959,plain]d-> "outputConfig\n1" OutputConfig
...
+## Choosing solvers
+
+`--abstraction-solver` / `--refinement-solver` (and the portfolio's own choices) are not
+interchangeable, because not every solver supports every theory the frontend may pick.
+
+**Bitvector arithmetic plus interpolation requires MathSAT.** The arithmetic is chosen from the
+program: a source using bitwise operators, bitfields or a byte-addressed union resolves to bitvectors
+even under `--arithmetic efficient`. Z3-legacy cannot interpolate over bitvectors, so any CEGAR
+configuration whose refinement is interpolation-based (`*_ITP`) then fails with
+`theory not supported by interpolation or bad proof`. Pin `mathsat:5.6.10` for both solvers in that
```
mondokm: I wouldn't refer to a specific mathsat version (`mathsat:5.6.10`) like this here. Also maybe the whole comment can go.

subprojects/common/common/src/test/java/hu/bme/mit/theta/common/logging/LoggerFormatSafetyTest.java
mondokm, line 1: This test seems like a debugging artifact that is not needed anymore.

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/anytype/RefExpr.java
```diff
@@ -49,7 +50,19 @@ public DeclType getType() {
 
     @Override
     public LitExpr<DeclType> eval(final Valuation val) {
-        return val.eval(decl).get();
+        // Name the unbound variable: `.get()` would throw a bare "No value present", which says
```
mondokm: Remove the comment

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/utils/TypeUtils.java
```diff
@@ -98,13 +99,17 @@ public static <T extends Type> Expr<T> cast(final Expr<?> expr, final T type) {
             final Expr<T> result = (Expr<T>) expr;
             return result;
         } else {
+            // Types first, expression last: an expression pretty-prints over many lines, so
```
mondokm: Remove the comment

mondokm (dismissed):
I flagged a few things as well, mostly:
- I agree with @csanadtelbisz that the excessive commenting is very annoying, 90% of the comments are not needed at all.
- There are some "Claude workflow artifacts", tests that were created for debugging, benchmarking workflow descriptions.

subprojects/common/analysis/src/main/java/hu/bme/mit/theta/analysis/ptr/PtrPrec.kt
leventeBajczi: resolved in 501a87a33c

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
leventeBajczi: resolved in 7d7bbd6993

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaOcChecker.kt
leventeBajczi: resolved in ed446e963c

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
leventeBajczi: resolved in 7d7bbd6993

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
leventeBajczi: resolved in 7d7bbd6993 -- it ended up as `MemoryTypeKey.sanitizedName`, a member property of the shared key class extracted for your other comment, rather than a member of `XcfaToEventGraph`.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
leventeBajczi: resolved in 7d7bbd6993

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
leventeBajczi: resolved in 7d7bbd6993 -- `releasableFrom`, `releasedEarly` and the debug block are gone and the traversal reports a stalled wait list again. You were right that it is the unused-loc removal's job: `UnrollPass` now runs it itself (ed446e963c), so the unreachable copies and their edges are gone before the event graph is built.

subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/params/XcfaConfig.kt
leventeBajczi: resolved in ed446e963c -- `--force-unroll-recursion <Int>`, defaulting to -1 (off), with `UnrollPass.UNROLL_RECURSION_LIMIT` beside `FORCE_UNROLL_LIMIT`. The OC checker passes its current escalation bound to the pass explicitly, so raising that bound still re-expands the recursion one level deeper.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/Builders.kt
leventeBajczi: resolved in a68847a628 -- the check moved into `ProcedurePass.runChecked`, which every call site (the two `optimize` loops plus the ad-hoc `pass.run` calls in FetchExecuteWriteback, OverflowDetectionPass, PthreadArrayHandleUnrollPass, SimplifyExprsPass, ReferenceElimination and UnrollPass) now uses. `run` itself stays public: an interface method cannot be made final in Kotlin.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/AllocaFunctionPass.kt
leventeBajczi: resolved in 9e39a88392 -- `mallocVar`, `ensureMallocVar`, `firstAllocationRetType` and the static-base-limit key are now in `xcfa/utils/AllocationUtils.kt`.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/AllocaFunctionPass.kt
leventeBajczi: resolved in 9e39a88392 -- written down on `POINTER_BASE_CLASSES` in `xcfa/utils/AllocationUtils.kt`, which is also the constant the four sites now share instead of a literal 3. There are three classes: `3k+0` heap (malloc/calloc), `3k+1` static objects and alloca, `3k+2` address-taken locals (ReferenceElimination). They cannot overlap: both runtime kinds draw from the single `__malloc` counter, which advances by 3, and the counter is seeded past the frontend's last compile-time base, which is the only class with two producers.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/CallocFunctionPass.kt
leventeBajczi: resolved in 9e39a88392

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/DataRaceToReachabilityPass.kt
leventeBajczi: resolved in 3a49945c9c and 1b5b0232e2 -- the per-variable read/write flags are counters now (increment on access, decrement on release; the `== 0` assertion is unchanged). One thing that had to come with it: at a branching location the edge into the branch flags the accesses of *every* branch condition while each branch cleared only its own, which a flag hid and a counter would not, so the aggregate is now collected up front and every branch clears all of it. The dereference flags are left alone -- they hold the address being accessed, not a count, so increment/decrement does not apply to them.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/DereferenceToArrayPass.kt
leventeBajczi: resolved in 7d7bbd6993 -- `hu.bme.mit.theta.xcfa.utils.MemoryTypeKey` (`arrayType`/`offsetType`/`elemType` plus `sanitizedName`) with a `Dereference.memoryTypeKey` accessor, used by both DereferenceToArrayPass and the OC event graph.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/UnrollPass.kt
leventeBajczi: resolved in ed446e963c for point 2: substitution is now refused when the loop variable is a global, where another thread could write it and a baked-in constant could hide a race or an out-of-bounds access.

1: SimplifyExprsPass cannot do this. It simplifies a label against a valuation it is given and has no dataflow of its own, and the loop variable's value differs per copy -- only the unroller knows which iteration a copy is. Deriving it afterwards would need a real constant-propagation pass over the unrolled CFA; that is worth having, but it is not a widening of the expression simplifier.

3: because it changes the model everywhere, not just where it is needed. It is on for PthreadArrayHandleUnrollPass, which needs `&t[i]` to be a constant index before CLibraryFunctionsPass reads the handle; turning it on for the ordinary unroll would need a benchmark campaign rather than an argument.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/UnrollPass.kt
leventeBajczi: resolved in ed446e963c

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/ProcedurePassManager.kt
leventeBajczi: resolved in ed446e963c

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/README.md
leventeBajczi: resolved in 3a49945c9c -- rewritten as a table of the actual CPasses pipeline, phase by phase with every pass in it, the passes that are backend-specific rather than in the pipeline, and the rules the passes follow. No changelist left.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/ReallocFunctionPass.kt
leventeBajczi: resolved in 9e39a88392 -- folded into MallocFunctionPass, which handles both shapes of the same rewrite; ReallocFunctionPass is gone.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/TypeRange.kt
leventeBajczi: resolved in 9e39a88392 -- `withinTypeRange` moved into `xcfa/utils/TypeDefaults.kt`, next to `Type.defaultValue`.

CLAUDE.md
leventeBajczi: resolved in 1f42a7277c -- the benchmark-launch sections are gone. I kept the file with the repo-level build/verification loop plus the short-comments rule @csanadtelbisz asked for in the review body

.gitignore
leventeBajczi: resolved in 1f42a7277c

subprojects/xcfa/xcfa/src/test/java/hu/bme/mit/theta/xcfa/passes/UnresolvedInvokeToHavocTest.kt
leventeBajczi: resolved in 1f42a7277c

subprojects/xcfa/xcfa-cli/README.md
leventeBajczi: resolved in 1f42a7277c -- the whole section is gone.

subprojects/common/common/src/test/java/hu/bme/mit/theta/common/logging/LoggerFormatSafetyTest.java
leventeBajczi: resolved in 1f42a7277c

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/type/anytype/RefExpr.java
leventeBajczi: resolved in 1f42a7277c -- comment removed; the named-declaration error message stays.

subprojects/common/core/src/main/java/hu/bme/mit/theta/core/utils/TypeUtils.java
leventeBajczi: resolved in 1f42a7277c

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/Builders.kt
leventeBajczi: Yes, we rely on the instances. `XcfaLocation` is a data class over (name, initial, final, error, metadata), but `incomingEdges`/`outgoingEdges` are mutable per-instance state that `equals` does not cover, so two equal locations are not interchangeable.

The case it catches: a pass constructs `XcfaLocation("l", metadata = m)` for a name the procedure already holds. `addLoc` is a silent no-op for it, so `locs` keeps the original instance, while an edge built from the new one satisfies `addEdge`'s `source in locs` by equality and is nevertheless absent from the registered instance's adjacency sets -- invisible to every traversal that walks adjacency, and re-pointed onto the registered instance later by `XcfaProcedure.deepCopy`, which resolves endpoints through an equality-keyed map. An equality-based check here cannot see any of that.

Agreed that making locations compare by identity is the better fix, but it changes deepCopy, the gson adapters and every location-keyed map, so I would rather not do it inside this PR. Happy to open it separately if you want it.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/XCFA.kt
leventeBajczi: `pointsToAtomic` sits on `XcfaGlobalVar`, so it can only answer for a live *global pointer variable*. What the memory-access check has to ask is whether the accessed *cell* is `_Atomic`: a field of a global struct, an element of an `_Atomic` array, or an address-taken `_Atomic` local whose pointer is a bare base-id literal by then. That is a property of the (base id, offset) pair, and there is no model element keyed by it today -- the registry lives in ParseContext because that is where the base ids are minted.

So the two are not redundant: the model field covers the one case that has a home in the model, and the rest still has to be looked up. Moving the per-cell registry into the XCFA (or into the type, as you suggest on the other thread) is the right fix and I agree with it, but it reaches across the frontend, the model and both race checkers, so I have left it out of this PR rather than half-doing it.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/LibraryStubsPass.kt
leventeBajczi: Checked, and both do add something, so I have left them where they are.

`printf`: CLibraryFunctionsPass materialises every variadic argument as an assignment into a `__printf_arg_*` variable, which keeps the *read* of those arguments in the model -- so a data race on `printf("%d", shared)` is still visible. A LibraryStubsPass entry would be `Writes.None`, whose whole expansion is a havoc of the return slot: the argument expressions disappear along with the InvokeLabel.

`scanf`: CLibraryFunctionsPass runs in the first pass group, before ReferenceElimination, so `&x` is still a `Reference` and it can havoc the variable `x` directly, for every argument. LibraryStubsPass runs long after, and can only write through the pointer into memory cells -- and when the extent is not known it fills `DEFAULT_FILL_CELLS` (1) of them. That is strictly weaker here.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/WitnessOptimizer.kt
leventeBajczi: They are in different Gradle modules and the dependency runs the wrong way: `WitnessOptimizer` is in `theta-xcfa` and is used from `theta-xcfa-analysis` (the OC checker builds its per-thread procedures with it), while `ApplyWitnessPass` is in `theta-xcfa-cli`, which depends on both. Merging them would make `theta-xcfa-analysis` depend on `theta-xcfa-cli`. Happy to move them together if the OC use can be relocated, but that is a bigger change than this comment.

leventeBajczi:
Thanks for your reviews, please check if I've addressed of them them properly, and resolve the threads if so.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
csanadtelbisz: Still not satisfied:) I commit my suggestions, that is the easiest.

subprojects/xcfa/xcfa-cli/src/main/java/hu/bme/mit/theta/xcfa/cli/params/XcfaConfig.kt
csanadtelbisz: I renamed the parameters as `always*` indicated a Boolean type parameter. Otherwise good, thank you.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/Builders.kt
csanadtelbisz: Moved `runChecked` to the ProcedurePass interface as there is no point in having it as an extension method.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/Builders.kt
csanadtelbisz: Oh, indeed, I see. We should probably clean up this in a later PR (e.g., having a separate XcfaLocationBuilder with mutable in and out edges and a pure immutable XcfaLocation in the Xcfa).

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/model/XCFA.kt
csanadtelbisz: Okay, let's keep it this way in this PR, as this should not be a blocking issue. But I create a new issue for this, and let's do it for in this year's svcomp iteration.
See #543.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/XcfaDataRaceCheck.kt
csanadtelbisz: Resolving this thread now, see #543.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/AllocaFunctionPass.kt
csanadtelbisz: Thank you, this is a very useful consolidation of our pointer space (I should say our 5D chess game).

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/LibraryStubsPass.kt
csanadtelbisz: This raises the question whether all other similar methods in LibraryStubsPass have a good place there:
- Do other printf-like methods that read some variables loose the read access and potentially hide a data race? If so, we should handle all such methods in LibraryStubsPass as we do with printf (and then, we could move printf there).
- Other scanf-like methods can potentially also introduce references, right? So they should also be before ReferenceElimination. As we already have a ReferenceElimination step in the procedure pass manager after function inlining, we could just move librarystubspass between these two steps. Is there any other dependency of LibraryStubsPass that must precede it in the pass order?

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/README.md
csanadtelbisz: It would be useful to have a list of "dependencies" between the passes in this README: maybe as a new dependencies column. E.g., for the function inlining or function pointer pass (whichever, I cannot really follow) a ReferenceElimination must occur. Or we must have a simplify expressions pass after unroll. Or LibraryStubsPass must be after function inlining, etc.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/WitnessOptimizer.kt
csanadtelbisz: Ah, okay.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/utils/AtomicAccessUtils.kt
csanadtelbisz: #543.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/UnrollPass.kt
csanadtelbisz: 1. SimplifyExprsPass has a basic constant propagation of its own. After unrolling, the value of the loop var should be propagated trivially. At least that is my expectation.

3. Sure, but if we do not have any argument against it, then we should do the change and test it with a benchmark campaign. But only after deciding point 1.

subprojects/xcfa/xcfa-analysis/src/main/java/hu/bme/mit/theta/xcfa/analysis/oc/XcfaToEventGraph.kt
csanadtelbisz: #544

csanadtelbisz:
Most of my comments are resolved, but there are still some open issues/questions.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/LibraryStubsPass.kt
leventeBajczi: resolved in 486c893dfe, for both halves of the question -- and yes, the read access was being lost.

Reproduced first. Two programs identical except for how one thread writes the shared global:

- `g = 7` racing another write -> `Unsafe`, race reported
- `fscanf(0, "%d", &g)` racing the same write -> `Safe`, **race missed**

and for the read side, `fprintf(0, "%d", shared)` racing a write to `shared` -> `Safe`, missed.

Two causes:

1. `LibraryStubsPass` ran after `DataRaceToReachabilityPass` and `MemsafetyPass`, so the writes it emits for a call's pointer arguments were never instrumented. It now runs at the end of the first pass group. Note the slot you suggested (after inlining, before the following `ReferenceElimination`) does **not** work -- I tried it and the write case stayed `Safe`. By then the first `ReferenceElimination` has folded `&g` into a bare base id, `CComplexType.getType` falls back to guessing from the SMT sort and answers `CSignedInt`, `pointeeCellType` returns null, and the stub silently models no write at all. The pass has to see `&g` while it is still a reference. Running earlier also computes `defined` while every procedure is still present.
2. Replacing the `InvokeLabel` dropped the argument expressions with it. Each argument is now materialised into a fresh variable first, as `CLibraryFunctionsPass` does for `printf`.

Only `UnresolvedInvokeToHavocPass` needs to be last, so it stays where it was.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/UnrollPass.kt
leventeBajczi: **Correction: my earlier reply was wrong.** I said `SimplifyExprsPass` "has no dataflow of its own". It does -- `mergeIncomingValuations` iterated to a fixpoint, plus `separateConstAndModifiedVars`. You were right.

Verified with `substituteLoopVar = false`: `for (int i = 0; i < 3; i++) t[i] = i + 10;` comes out of the pipeline fully propagated as `(memassign (deref 1 0) 10)`, `(deref 1 1) 11`, `(deref 1 2) 12`. So for the ordinary case the flag is redundant, exactly as you expected.

It is still needed in one place: `PthreadArrayHandleUnrollPass` runs in pass group 1, two groups before the first `SimplifyExprsPass`. With the flag off, `pthread_create(&t[i], ...)` fails there with `Unsupported library parameter: non-constant dereference offset (main::for1::i)`.

So on your point 3 -- always substituting is not blocked by any argument I can find, but the flag cannot simply be deleted while that pass needs a constant offset before any simplification has run. Two options: keep the flag for that one caller, or move the constant folding it needs earlier. Happy to do either; the second is the cleaner fix and wants a benchmark campaign to confirm, as you say.

No code change in this push beyond the global-loop-variable guard already in ed446e963c.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/README.md
leventeBajczi: resolved in 9303e385c0 -- added a dependencies column.

subprojects/xcfa/xcfa-cli/canaries/run_fixtures.sh
leventeBajczi: resolved in a38c97eb89 -- `FixtureSuiteTest` plus a `fixtureTest` Gradle task, alongside the existing `canaryTest`. One dynamic test per row of `fixtures/fixtures.tsv`; 75/75 green.

When the sv-benchmarks checkout is missing it fails with a message naming the expected location, and it never fetches it.

csanadtelbisz: So do we still need the shell script or is it completely replaced by the new gradle task? If so, please remove the shell script.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/LibraryStubsPass.kt
csanadtelbisz: The original question still remains: could not we move the printf and scanf implementations from CLibraryFunctionsPass to LibraryStubsPass now? Is there still a reason for keeping them apart?

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/UnrollPass.kt
leventeBajczi: **1. Not sound for a non-local loop variable** -- verified, not assumed. Global `i`, another thread sets it past an array bound, `a[i]` in the loop:

- guard removed: `Safe` (the out-of-bounds is gone, because the copies read `a[0]`/`a[1]`)
- guard in place: refused with `non-constant dereference offset`

So the guard in ed446e963c is right, but note what it costs: refusing to substitute does not make the loop analysable, it makes `&t[i]` non-constant and the task errors out. On a *racy* global loop variable the no-guard build actually answers `Unsafe` correctly, so the guard trades one correct answer for one avoided wrong answer. Net positive under the scoring, but it is a refusal, not a fix.

(Substitution replaces only reads; the loop's own writes to `i` remain, so races on `i` itself stay visible. The unsoundness is the memory-safety one.)

**2. Always substituting: neutral locally.** Enabled it for the main `UnrollPass` and the canary suite is 268/268, no regression. But those are parse-mode, so they can only show nothing broke -- not that anything improved. Deciding this needs a verdict-level benchmark campaign; I have not run one.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/LibraryStubsPass.kt
leventeBajczi: resolved in a058f6969b -- you were right, and my earlier reason no longer held: `LibraryStubsPass` now materialises argument reads *and* runs before `ReferenceElimination`, which were exactly the two things keeping them apart. Both moved.

It also fixes a split I had not noticed: `scanf` was in `CLibraryFunctionsPass` while `__isoc99_scanf` -- what glibc renames it to, so what actually appears in preprocessed sources -- was already a stub.

Verified `scanf("%d", &x)` still havocs `x`. `PassTests` asserted the `__printf_arg_*` variables, so that case drops printf/scanf and `LibraryStubsTest` gains equivalent coverage.

subprojects/xcfa/xcfa-cli/canaries/run_fixtures.sh
leventeBajczi: resolved in a058f6969b -- removed, along with its call from `run_canaries.sh` and the dead fixture parsing in `CanarySuiteTest`.

One consequence worth stating: a bare `./run_canaries.sh` no longer runs the fixtures. `gradle :theta-xcfa-cli:canaryTest` still does, since it depends on `fixtureTest`.

csanadtelbisz: And `run_canaries.sh`, couldn't we remove that as well and replace with a gradle task? Sorry for asking it separately, I hoped the LLM will get the clue and remove all shell scripts.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/UnrollPass.kt
leventeBajczi: Concrete tasks for point 1. All three are the plain `pthread_t t[N]; for (i) pthread_create(&t[i], ...)` shape:

- `pthread-deagle/airline-10.i` — `pthread_t salethreads[10]`, create+join loops
- `pthread-complex/workstealqueue_mutex-1.i` — `pthread_t handles[2]`
- `pthread-complex/safestack_relacy.i` — `pthread_t threads[3]`

Flipping only `substituteLoopVar` to `false` in `PthreadArrayHandleUnrollPass` (nothing else changed), parse-mode `--backend NONE`:

| | with | without |
|---|---|---|
| airline-10 | `ParsingResult Success` | `non-constant dereference offset ((bvpos main::for8::i))` |
| workstealqueue_mutex-1 | `ParsingResult Success` | `non-constant dereference offset ((bvpos main::i))` |
| safestack_relacy | `ParsingResult Success` | `non-constant dereference offset ((bvpos main::i))` |

I swept all 97 sv-benchmarks tasks that create or join through an array element: 9 parse, and all 9 flip to that same error without the flag (also `airline-{5,15,20,25}`, `workstealqueue_mutex-2`, `safestack_relacy_racy`). So it is not a corner case, it is every task in the family that works at all today.

The reason `SimplifyExprsPass` cannot cover it is ordering, not capability. In `CPasses` the handle has to resolve at `PthreadArrayHandleUnrollPass` → `CLibraryFunctionsPass` (lines 50-51), while `SimplifyExprsPass` and the ordinary `UnrollPass` are lines 69-70. At line 51 the loop is still rolled, so there is no per-iteration constant for any propagation to find — your expectation holds for a propagator running after unrolling, but nothing has unrolled yet at that point. Moving `CLibraryFunctionsPass` later is not an option either: `ReferenceElimination` in between rewrites `&t[i]` into base/offset and the handle identity is gone.

(Sweep also shows 69 of the 97 still fail with the same message *with* the flag on — different shapes it does not reach: `indexer.i` indexes a mutex array by a function parameter, `sigma.i` uses a malloc'd `pthread_t *`, `reorder_2.i` a VLA sized from `argv`. Separate issue.)

leventeBajczi: Tried the obvious version of this: `SimplifyExprsPass` inserted between `PthreadArrayHandleUnrollPass` and `CLibraryFunctionsPass`, `substituteLoopVar` deleted outright.

It does fold the handles — your expectation about the propagation was right, my earlier "only the unroller knows which iteration a copy is" was wrong. All 9 sv-benchmarks array-handle tasks parse, and the full 97-task sweep is unchanged.

But it changes a verdict: the `memcpy_struct_cells` fixture goes **SAFE → UNSAFE**. `&d` on an aggregate is not a `Reference` — the frontend gives the global its storage with `d := 2` (`FrontendXcfaBuilder`), so `&d` *is* the variable holding the base id, and constant propagation folds it to the literal. That is semantically fine but breaks every later pass that matches on the base *variable*: the copy was lowered against an object `ReferenceElimination` never allocated.

I had already patched exactly this conflict for `getParam` (resolving a folded base id back to its global). `MemoryFunctionsPass` hitting the same wall is what convinced me that patching it per-pass is the wrong shape. The principled fix is to stop the simplifier substituting object-base globals at all, which needs the frontend to record which globals those are — a bigger change than this PR, and your call whether it is worth opening.

So `substituteLoopVar` stays for now, and point 3 has a concrete argument against it rather than just an absence of one.

Two real bugs fell out of the attempt, fixed in 33f30ce952 (both latent on master, independent of this decision):
- `XcfaLabel.simplify` rebuilt `InvokeLabel`/`StartLabel` arguments without their `cType`, so `AtomicFunctionsPass` refused its own pointer argument as a non-pointer. The statement branches beside it already re-attached it.
- `ExprSimplifier` had no `Reference` case (it was commented out), so it recursed into the operand and turned `&x` into `&1`. Reachable today via `substituteLoopVar`, which runs before `ReferenceElimination` while references still exist.

Gates: 75/75 fixtures, 267/268 canaries (the one failure is an OOM kill that reproduces identically on a clean HEAD build), core 679 tests green.

leventeBajczi: Replacing my earlier replies on this thread with one summary, since the investigation superseded most of them.

**Outcome: the parameter stays, and e5596087e8 documents why at the parameter and in the pass README.**

**2 (soundness).** Resolved in ed446e963c: substitution is refused when the loop variable is a global, where another thread could write it and a baked-in constant could hide a race or an out-of-bounds access. Verified both directions on a repro.

**1 and 3.** You were right that `SimplifyExprsPass` can do this — my earlier claim that it has no dataflow was wrong. It folds `&t[i]` correctly once it runs after the unroll. What blocks the cleanup is pass ordering, and after prototyping it end to end I do not think there is a valid one:

- Simplification must run **after** `ReferenceElimination`. Otherwise it folds the variable naming an object into that object's base id, and every later pass that matches on the variable stops recognising the object. Concretely, `memcpy` from a global struct then copies an object whose initialisers were re-based elsewhere, and the `memcpy_struct_cells` fixture flips SAFE → UNSAFE. (Separately, simplification used to rewrite `&x` into `&1`; that one is now guarded in `ExprSimplifier`, 33f30ce952.)
- `ReferenceElimination` must run **after** `AtomicFunctionsPass`, `LibraryStubsPass` and `CLibraryFunctionsPass`, which all read `&x` arguments as references.
- `CLibraryFunctionsPass` cannot move after it at all. An address-taken local is re-based onto a *runtime* counter (`ptrVar += step`), so there is no static base id — a thread handle loses the identity that matches a create to its join. I measured this: with the reordering, every one of 25 sampled ordinary `pthread_t t1, t2;` tasks stops parsing. Only tasks whose handles are globals survive.

So the cycle is real: unroll → simplify wants RE first, RE wants the handle passes first, the handle passes want the unroll first.

It is still doable — it means giving pthread handles an identity that survives re-basing (a stable per-object id recorded where the base is minted), so `CLibraryFunctionsPass` can run after `ReferenceElimination`. That is a design change rather than a reordering, and felt out of scope here. Happy to open a separate issue if you think it is worth it.

Two real bugs fell out of the attempt and are fixed in 33f30ce952, both independent of this decision: `XcfaLabel.simplify` dropped `cType` when rebuilding `InvokeLabel`/`StartLabel` arguments, and `ExprSimplifier` had no `Reference` case so it recursed into the operand and turned `&x` into `&1`.

subprojects/xcfa/xcfa-cli/canaries/run_fixtures.sh
csanadtelbisz: Ping.

csanadtelbisz (approved):
Thank you for the very much needed changes! All of my concerns are resolved - or issues have been created for some of the non-blocking ones.

mondokm (approved):

leventeBajczi:
## Run 110 — full SV-COMP, Theta 8.0.0, `--portfolio COMPLEX27`, 900s / 15GB / 2 cores

36531 distinct task+property pairs (72103 raw runs; this benchmark definition lists each task twice, figures below are deduplicated).

| | count | % |
| --- | --- | --- |
| correct | 14018 | 38.4 |
| &nbsp;&nbsp;true / false | 8392 / 5626 | |
| wrong | 102 | 0.3 |
| &nbsp;&nbsp;missed bugs / false alarms | 18 / 84 | |
| timeout | 17082 | 46.8 |
| out of memory | 2821 | 7.7 |
| tool error | 2103 | 5.8 |
| unknown | 397 | 1.1 |
| **score** | **20490** | |

| property | n | correct | wrong | missed | timeout | oom | error | score |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| unreach-call | 18738 | 4763 | 22 | 11 | 10357 | 2339 | 1257 | 7853 |
| no-overflow | 9027 | 5656 | 38 | 1 | 2980 | 118 | 235 | 8186 |
| valid-memsafety | 5248 | 1811 | 33 | 4 | 2585 | 28 | 386 | 1793 |
| termination | 2395 | 1032 | 2 | 0 | 1010 | 332 | 19 | 1451 |
| no-data-race | 1030 | 713 | 1 | 0 | 121 | 4 | 191 | 1292 |
| valid-memcleanup | 93 | 43 | 6 | 2 | 29 | 0 | 15 | -85 |

Against run 102 (7.3.0, identical config and CPU model): score 22536 -> 20490, correct -848, timeout +1028, tool error -234, missed bugs 26 -> 18. Filed as #564 with task lists.

## Frontend failures — 1893, filed per family

| # | count / distinct tasks | family |
| --- | --- | --- |
| #553 | 417 / 366 | `No such variable or macro` |
| #554 | 287 / 278 | floating-point member of a byte-addressed union |
| #555 | 223 / 109 | non-constant dereference offset in a pthread handle |
| #556 | 177 / 138 | `__strcpy_index_var` is Int where a bitvector is expected |
| #557 | 156 / 142 | inlining arity mismatch |
| #558 | 155 / 61 | local mutex handles |
| #559 | 86 / 86 | assignment LHS is a BvAdd/BvPos expression |
| #560 | 39 / 36 | `TypeUtils.cast` type mismatch |
| #561 | 35 / 15 | pthread handle base is not a reference variable |
| #562 | 27 / 21 | `typeof` over an undeterminable expression |
| #563 | 10 / 10 | `fesetround` with a non-default rounding mode |

281 further failures over 242 tasks are not yet grouped.

#555, #558 and #561 (413 failures) are the `CLibraryFunctionsPass.getParam` handle-identity limitation described in the `substituteLoopVar` thread. #556 is new in this build: the variable comes from the `strcpy` lowering merged in #536.

PR #531, opened 2026-08-30
Unified precise time measurements

s0mark:
Extends the `cgroups`-based precise CPU-time measurements introduced by #468 in `MddChecker` to the whole of Theta. Keeps the previously used Google stopwatch as a fallback if `cgroups` is not available.


mondokm (approved):
Looks good to me, I agree we should start using this elsewhere too. What is the reason for the Stopwatch.kt -> Stopwatch.java change, is it to make it easier to use from java?

s0mark:
> What is the reason for the Stopwatch.kt -> Stopwatch.java change, is it to make it easier to use from java?

Exactly, so that static method accesses from Java don't have to go through the `Companion`. Since this is how they're instantiated, I figured it was worth the simplification.

PR #566, opened 2026-09-10
strcpy type fix

csanadtelbisz:
Closes #556.
Probably the actual culprit was ReferenceElimination trying to assign an arbitrary type to an offset (the strcpy index counter has its own right to have a different Int type). Nevertheless, it was easier to patch CLibraryFunctionsPass to use the type of the copied dereference which actually matches the type forced by reference elimination.

PR #567, opened 2026-09-10
Drop unread string literal initialization, and copy strcpy's terminator

leventeBajczi:
Two frontend/pass fixes found while bisecting the run 110 regression (#564).

## 1. String literal initialization nothing reads

`4724647fe7` gave every string literal its storage and filled it one cell per character. In the string-heavy benchmarks that is most of the model and none of it is reachable by the property: the pointer is compared against null and handed to a library stub, and no cell is ever read back.

`StringLiteralInitPass` drops that initialization for a literal whose contents nothing reads. A literal is kept if any dereference **reads** one of its cells, or if the pointer is **stored into memory** (from where it could be loaded again and dereferenced through a base the pass cannot recognize). Passing it to a call or assigning it to a variable is not enough on its own: the address is followed through assignments to a fixpoint, and any variable used as a dereference base keeps everything it can hold. It runs after inlining, where a read through a callee's parameter has become a read through the literal.

On `Juliet_Test/CWE476_NULL_Pointer_Dereference__int64_t_31_good`:

| | locations | memory writes | KIND |
| --- | --- | --- | --- |
| `1051185d92` (before `4724647fe7`) | 13 | 1 | Safe 2s |
| `4724647fe7` | 147 | 65 | timeout >150s |
| this branch | 19 | 1 | Safe 19s |

803 of the 1098 `correct -> timeout` tasks in #564 are `Juliet_Test`.

## 2. strcpy did not copy the terminating null byte

The loop copies while `src[i] != 0` and exits on `src[i] == 0`, but never wrote that zero to the destination, so the destination was left unterminated:

```c
char dst[4];
strcpy(dst, "AB");
if (dst[0] != 'A' || dst[1] != 'B') reach_error();  // Safe: the characters copy
if (dst[2] != 0) reach_error();                     // Unsafe: the terminator is missing
```

The terminator is now written on the exit edge. It writes `sourceDeref`, which the assume has just constrained to zero, so it needs no literal of its own and composes with the typing change in #566.

Note for #566: that PR turns 177 tasks from `Frontend failed!` into real runs. Until this is in, some of those runs report a wrong verdict rather than an error.

## Guards

- `string_literal_read.c` -- a literal whose characters are read keeps its initialization. Fails (UNSAFE) against an earlier version of this pass that only looked at the dereference root.
- `strcpy_terminator.c` -- fails (UNSAFE) without the terminator write.

77/77 fixtures, 268/268 canaries, `theta-xcfa` and `theta-core` unit tests, spotless clean.

### DIFF
--- a/subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/CLibraryFunctionsPass.kt
+++ b/subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/CLibraryFunctionsPass.kt
@@ -156,7 +156,11 @@ class CLibraryFunctionsPass(val parseContext: ParseContext) : ProcedurePass {
                 builder.addEdge(copyEdge)
 
                 val exitAssume = StmtLabel(AssumeStmt.of(EqExpr.create2(sourceDeref, Int(0))))
-                val exitLabel = SequenceLabel(listOf(exitAssume))
+                // strcpy copies the terminating null byte as well; leaving it out left the
+                // destination unterminated, so reading the byte past the copied characters gave
+                // whatever was there before.
+                val copyTerminator = StmtLabel(MemoryAssignStmt.of(targetDeref, sourceDeref))
+                val exitLabel = SequenceLabel(listOf(exitAssume, copyTerminator))
                 val exitEdge = XcfaEdge(loc, target, exitLabel, metadata)
                 builder.addEdge(exitEdge)
               }
--- a/subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/StringLiteralInitPass.kt
+++ b/subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/StringLiteralInitPass.kt
@@ -1,0 +1,147 @@
+/*
+ *  Copyright 2026 Budapest University of Technology and Economics
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package hu.bme.mit.theta.xcfa.passes
+
+import hu.bme.mit.theta.core.decl.VarDecl
+import hu.bme.mit.theta.core.stmt.AssignStmt
+import hu.bme.mit.theta.core.stmt.MemoryAssignStmt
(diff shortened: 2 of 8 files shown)


subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/CLibraryFunctionsPass.kt
```diff
@@ -167,7 +167,11 @@ class CLibraryFunctionsPass(val parseContext: ParseContext) : ProcedurePass {
                   StmtLabel(
                     AssumeStmt.of(EqExpr.create2(sourceDeref, copySource.type.integerOf(0)))
                   )
-                val exitLabel = SequenceLabel(listOf(exitAssume))
+                // strcpy copies the terminating null byte as well; leaving it out left the
+                // destination unterminated, so reading the byte past the copied characters gave
+                // whatever was there before. `sourceDeref` is the zero the assume just pinned.
+                val copyTerminator = StmtLabel(MemoryAssignStmt.of(targetDeref, sourceDeref))
```
csanadtelbisz, line 173: Why not reuse the `copyCurrent` variable for copying the terminator? It is exactly the same statement label. Please also reduce the comment, a single comment line can easily explain that copying the terminating null by happens here.

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/StringLiteralInitPass.kt
```diff
@@ -0,0 +1,143 @@
...
+        }
+      }
+      val keep = readDirectly + storedToMemory + derefBases.flatMap { mayHold[it].orEmpty() }
+      return initialized - keep
+    }
+  }
+
+  private fun XcfaLabel.withoutInitOf(dead: Set<VarDecl<*>>): XcfaLabel =
+    when (this) {
+      is SequenceLabel -> SequenceLabel(labels.map { it.withoutInitOf(dead) }, metadata)
```
csanadtelbisz, line 129: I know we do not really use NondetLabel, but why is it left out here?

subprojects/xcfa/xcfa/src/main/java/hu/bme/mit/theta/xcfa/passes/StringLiteralInitPass.kt
```diff
@@ -0,0 +1,143 @@
...
+    val initialized = mutableSetOf<VarDecl<*>>()
+    val readDirectly = mutableSetOf<VarDecl<*>>()
+    val storedToMemory = mutableSetOf<VarDecl<*>>()
+    val derefBases = mutableSetOf<VarDecl<*>>()
+    val assignments = mutableListOf<Pair<VarDecl<*>, List<VarDecl<*>>>>()
+
+    fun visit(label: XcfaLabel) {
+      when (label) {
+        is NondetLabel -> label.labels.forEach(::visit)
+        is SequenceLabel -> label.labels.forEach(::visit)
```
csanadtelbisz, line 87: `label.dereferencesWithAccessType` also works for `NondetLabel` and `SequenceLabel`. Why implementing it separately?

csanadtelbisz (changes requested):

PR #568, opened 2026-09-10
Unroll a loop whose counter is initialized alongside another variable

leventeBajczi:
`UnrollPass` looks backwards from the loop start for the edge that initializes the loop variable, and required that edge to touch **exactly one** variable:

```kotlin
if (vars[loopVar].isWritten) {
    if (vars.size > 1) return@run null   // gives up on the loop entirely
    loopVarInit = inEdge
```

A global counter is initialized on the same edge as the allocation counter, so the check fires and the loop is never unrolled. Instrumented on `nla-digbench-scaling/geo1-ll_unwindbound1`:

```
DBGUNROLL loopVar=counter init edge writes it but touches 2 vars: [__malloc, counter]
```

The restriction is not needed. `count()` replays the whole edge from an empty state, so the other assignments are executed along the way, and a loop variable whose value stays unknown makes the count give up on its own.

`geo1-ll_unwindbound1` with `--backend KIND`:

| | |
| --- | --- |
| before | timeout (>150s) |
| after | Safe, 1-2s |

`cohencu-ll_unwindbound1` and `cohencu-ll_unwindbound10` also go from timeout to a verdict in seconds. These are part of the `correct -> timeout` set in #564 (32 tasks have this shape).

Guarded by a `PassTests` case rather than a fixture, since this is a performance property and a timing-based fixture would be flaky: the case asserts the loop *is* unrolled when its initialization shares an edge with an unrelated assignment. It fails (`[9] UnrollPass FAILED`) with the one-line change reverted.

268/268 canaries, 75/75 fixtures, `theta-xcfa` unit tests, spotless clean.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -86,7 +86,7 @@ buildscript {
 
 allprojects {
     group = "hu.bme.mit.theta"
-    version = "8.0.3"
+    version = "8.0.6"
 
     apply(from = rootDir.resolve("gradle/shared-with-buildSrc/mirrors.gradle.kts"))
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -86,7 +86,7 @@ buildscript {
 
 allprojects {
     group = "hu.bme.mit.theta"
-    version = "8.0.1"
+    version = "8.0.3"
 
     apply(from = rootDir.resolve("gradle/shared-with-buildSrc/mirrors.gradle.kts"))
 }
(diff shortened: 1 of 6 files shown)


leventeBajczi:
Added a second unroll fix to this PR, since it is the same pass and the same benchmark run.

`pthread-deagle/reorder_c11_*` (10 tasks, correct at 7.3.0) fail the frontend with
`Unsupported library parameter: non-constant dereference offset ((bvpos i))`. They index their
thread-handle array with a **file-scope** counter — `pthread_create(&setPool[i], …)` — and
`CLibraryFunctionsPass` keys a handle on a constant element index, so the counter has to reach the
unrolled copies as a constant.

The guard that blocked it refused substitution for *every* global. Its reasoning holds — another
thread writing the global would make a baked-in value wrong, hiding a race — so this commit narrows
it rather than removing it: substitute only when the loop sits in an init procedure (runs once) and
no other procedure writes the counter.

Guards: a fixture plus three unit tests, one per clause of the new condition. Each was checked
against the matching mutation —

| mutation | test that fails |
|---|---|
| restore the blanket guard | `aCounterNoOtherProcedureWritesIsSubstituted` |
| drop the guard entirely | `aCounterAnotherProcedureWritesIsNotSubstituted` |
| forget the init-procedure clause | `aCounterInAProcedureThatMayRunConcurrentlyIsNotSubstituted` |

The fixture fails with the frontend error on a build without the fix and returns SAFE with it.
`reorder_c11_bad-10` and `good-50` both verify to `false`, both expecting `false`.

Gate: canaryTest 268/268, fixtureTest 76/76, theta-xcfa unit tests 66/66, spotless and copyright clean.

